PR #25, opened 2023-07-25
multichar handling

### LINKED ISSUE #20: Support languages other than English

domino14:
This PR handles multi-character tiles. This can be both tiles such as `Ä` (non-ASCII), tiles such as `QU`, or a combination of both, such as `L·L`.

The PR does not distinguish between these cases. It treats user-visible tiles as char-arrays. Note that it doesn't care about encoding. One should provide distribution files (such as catalan.csv, for example) UTF8-encoded, for consistency, but the code doesn't actually care about the encoding. It just cares that the bytes match the input bytes coming in, via CGP strings for example. 

closes #20


core/src/klv.c
```diff
@@ -207,6 +212,9 @@ double get_leave_value(KLV *klv, Rack *leave) {
   if (leave->empty) {
     return 0.0;
   }
+  if (klv == NULL) {
```
jvc56, line 215: When do we expect this to be true?

core/src/letter_distribution.c
```diff
@@ -51,26 +51,33 @@ void load_letter_distribution(LetterDistribution *letter_distribution,
...
+    letter_distribution->machine_letter_to_human_readable_letter[i][0] = '\0';
+  }
 
   int machine_letter = 0;
   char line[100];
+  int max_tile_length = 0;
   while (fgets(line, sizeof(line), file)) {
-    // For now, we assume 1 char == 1 letter
     // This does not hold true for all languages
     // and will have to be updated.
```
jvc56, line 64: These also need to be deleted (it was not obvious they were part of the same comment).

core/src/letter_distribution.c
```diff
@@ -114,23 +118,66 @@ void load_letter_distribution(LetterDistribution *letter_distribution,
...
-      letter_distribution->human_readable_letter_to_machine_letter[r];
+                                        char *letter) {
+
+  for (int i = 0; i < MACHINE_LETTER_MAX_VALUE; i++) {
+    if (strcmp(letter_distribution->machine_letter_to_human_readable_letter[i],
+               letter) == 0) {
+      return i;
+    }
+  }
+  // It better find it
```
jvc56, line 138: Should we be aborting/asserting or otherwise fatally exiting here if we always expect to find the letter?

core/src/letter_distribution.c
```diff
@@ -114,23 +118,66 @@ void load_letter_distribution(LetterDistribution *letter_distribution,
...
+      if (j > num_bytes) {
+        continue;
+      }
+      // possible letter goes from index i to j. Search for it.
+      char possible_letter[MAX_LETTER_CHAR_LENGTH];
+      memcpy(possible_letter, str + i, j - i);
+      possible_letter[j - i] = '\0';
+      uint8_t ml = human_readable_letter_to_machine_letter(letter_distribution,
+                                                           possible_letter);
+      if (ml == INVALID_LETTER) {
```
jvc56, line 171: Same as above.

core/src/rack.c
```diff
@@ -70,18 +70,22 @@ int score_on_rack(LetterDistribution *letter_distribution, Rack *rack) {
 void set_rack_to_string(Rack *rack, const char *rack_string,
                         LetterDistribution *letter_distribution) {
   reset_rack(rack);
-  for (size_t i = 0; i < strlen(rack_string); i++) {
-    add_letter_to_rack(rack, human_readable_letter_to_machine_letter(
-                                 letter_distribution, rack_string[i]));
+
+  uint8_t mls[50]; // idk
```
jvc56, line 74: nit: this should probably use the RACK_SIZE constant in some way if I am reading this correctly, but isn't really necessary as long as it's big enough.

core/test/alphabet_print.c
```diff
@@ -6,6 +8,11 @@
 // Assumes english for now
```
jvc56, line 8: I think we can delete this comment now.

core/src/klv.c
domino14: for situations where we don't have leaves for a specific lexicon

core/src/letter_distribution.c
domino14: I'll delete this comment; it's actually ok to return INVALID_LETTER (see algo below)

core/src/letter_distribution.c
domino14: this is ok, as discussed in voice chat, it's an expected value.

core/src/rack.c
domino14: will change

PR #26, opened 2023-07-25
Basic UCGI implementation

domino14:
Implements basic UCGI commands:

`ucgi`

`position cgp <cgpstring>`

`go plies 5 stopcondition 99 threads 10`

`stop`

`exit`

change base to master once `multichar-handling` is merged

### DIFF
--- a/core/src/ucgi.c
+++ b/core/src/ucgi.c
@@ -5,6 +5,7 @@
 #include "game.h"
 #include "log.h"
 #include "sim.h"
+#include "ucgi.h"
 
 #define CMD_MAX 256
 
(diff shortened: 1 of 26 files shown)


core/src/ucgi.c
```diff
@@ -0,0 +1,198 @@
...
+void set_mode_stopped_callback() {
+  current_mode = MODE_STOPPED;
+  log_debug("setting current mode to stopped");
+}
+
+void load_position(const char *cgp, char *lexicon_name, char *ldname) {
+  char dist[50];
+  sprintf(dist, "data/letterdistributions/%s.csv", ldname);
+  char leaves[50] = "data/lexica/english.klv2";
+  char winpct[50] = "data/strategy/default_english/winpct.csv";
```
jvc56, line 39: Will the leaves and win tables be generalized to nonenglish in the future?

domino14: yes, we can do that in the future. In the `if` statement after these statements, we can overwrite these variables depending on what the lexicon is.

domino14:
Some rationale for stopping threads and this scary loop:

```c

  tc->status = THREAD_CONTROL_EXITING;

  // if we're here, this thread is about to stop.
  if (tc->thread_number == 0) {
  // Wait until all the other threads are done.
  waitfordone:
    for (int i = 0; i < tc->simmer->threads; i++) {
      if (tc->simmer->thread_control[i]->status != THREAD_CONTROL_EXITING) {
        // https://xkcd.com/292/
        goto waitfordone;
      }
    }
    log_debug("all threads about to exit");
    // Print out the bestplay, UCGI. (and the final rankings/data)
    if (tc->simmer->ucgi_mode == UCGI_MODE_ON) {
      print_ucgi_stats(tc->simmer, 1);
    }
    if (tc->simmer->endsim_callback != NULL) {
      tc->simmer->endsim_callback();
    }
  }
```

Note that this code is inside the `single_thread_simmer`. It waits for all the other threads to hit the top line `tc->status = THREAD_CONTROL_EXITING;`.

First, the reason for the loop is that we want to make sure no threads are simming any more data before we print out the best play and final stats. Yeah, `goto` isn't always great, but it's fairly easy to understand what it's doing here 😝 

Since the threads are already detached, all resources will be released once the threads exit.

Since the code is inside `single_thread_simmer`, we can't join thread 0. We could probably join the other threads instead of the waitfordone loop, but then we'd have to not `detach` the other threads (you can't join a detached thread). So we'd have thread 0 be detached but threads 1 through n would not be detached. It might be a little inelegant.

If instead we make it so no threads are detached, it was not clear to me where I would call `join` from. When we do a "stop" in ucgi, we would like that thread (the main thread, running the ucgi loop) to remain responsive. We could potentially call a function that calls `join` on all the threads from that ucgi loop, but that might make it briefly unresponsive. This could be a solution, though.

The function's body could look like 
```c
  for (int t = 0; t < simmer->threads; t++) {
    pthread_join(simmer->thread_control[t]->thread, NULL);
    simmer->thread_control[t]->status = THREAD_CONTROL_IDLE;
  }
```
We'd have to call it after we set the "stop" flags on the threads of course. Let me know what you think. Overall explicitly joining the thread is probably more elegant, but I don't think it's necessarily wrong to use detach here.

jvc56 (approved):

PR #35, opened 2023-08-16
Infer ucgi

jvc56:


core/src/sim.c
```diff
@@ -290,36 +277,116 @@ int handle_potential_stopping_condition(Simmer *simmer) {
...
+        game->gen->bag->last_tile_index + 1 +
+            game->players[1 - simmer->initial_player]->rack->number_of_letters,
+        plies % 2, is_multithreaded(simmer));
+    // reset to first state. we only need to restore one backup.
+    unplay_last_move(game);
+  }
+}
+
+void *simmer_worker(void *uncasted_simmer_worker) {
+  SimmerWorker *simmer_worker = (SimmerWorker *)uncasted_simmer_worker;
```
domino14, line 350: this cast is not technically necessary I learned. According to my copy of K&R it doesn't seem to be needed 😝

core/src/ucgi_print.c
```diff
@@ -0,0 +1,242 @@
...
+    uint64_t niters = play->equity_stat->cardinality;
+    store_move_ucgi(play->move, game->gen->board, move,
+                    game->gen->letter_distribution);
+
+    sprintf(stats_string + strlen(stats_string),
+            "info currmove %s sc %d wp %.3f wpe %.3f eq %.3f eqe %.3f it %lu "
+            "ig %d ",
+            move, play->move->score, wp_mean, wp_se, eq_mean, eq_se, niters,
+            play->ignore);
+    sprintf(stats_string + strlen(stats_string), "plies ");
```
domino14, line 221: can use return value of sprintf like ```c
  char coords[20];
  tp = coords;

  if (move->move_type == MOVE_TYPE_PLAY) {
    if (move->vertical) {
      tp += sprintf(tp, "%c", move->col_start + 'a');
      tp += sprintf(tp, "%d", move->row_start + 1);
    } else {
      tp += sprintf(tp, "%d", move->row_start + 1);
      tp += sprintf(tp, "%c", move->col_start + 'a');
    }
    ```

core/src/ucgi_print.c
```diff
@@ -0,0 +1,242 @@
...
+  if (number_of_tiles_played_or_exchanged == 0) {
+    number_of_tiles_played_or_exchanged =
+        (RACK_SIZE)-inference->initial_tiles_to_infer;
+  }
+  Game *game = inference->game;
+
+  // Create a transient stat to use the stat functions
+  Stat *letter_stat = create_stat();
+
+  char records_string[500000] = "";
```
domino14, line 128: 🙈

domino14 (changes requested):
couple small things

core/src/sim.c
jvc56: The compiler gives warnings if this isn't included.

PR #45, opened 2023-08-28
Add checkstop mutex

jvc56:


core/src/sim.c
```diff
@@ -247,9 +231,10 @@ int handle_potential_stopping_condition(Simmer *simmer) {
   SimmedPlay *tentative_winner = simmer->simmed_plays[0];
   double mu = tentative_winner->win_pct_stat->mean;
   double stderr = get_standard_error(tentative_winner->win_pct_stat, zval);
-  int new_ignored = 0;
+  int total_ignored = 0;
   for (int i = 1; i < simmer->num_simmed_plays; i++) {
```
domino14, line 235: if we're counting all ignored plays shouldn't this start at `i = 0` ?

core/src/sim.c
```diff
@@ -210,24 +212,6 @@ void ignore_play(SimmedPlay *sp, int lock) {
 }
 
 int handle_potential_stopping_condition(Simmer *simmer) {
-  if (simmer->num_simmed_plays < 2) {
-    return 1; // should stop
-  }
-  if (simmer->iteration_count >
-      (MAX_STOPPING_ITERATION_CT +
-       (simmer->max_plies * PER_PLY_STOPPING_SCALING))) {
```
domino14, line 218: since we removed this, is your thought that # iterations should just be a fixed per-invocation thing?

core/src/sim.c
```diff
@@ -210,24 +212,6 @@ void ignore_play(SimmedPlay *sp, int lock) {
...
-  int ignored_plays = 0;
-  for (int i = 0; i < simmer->num_simmed_plays; i++) {
-    if (simmer->simmed_plays[i]->ignore) {
-      ignored_plays++;
-    }
-  }
-  if (ignored_plays >= simmer->num_simmed_plays - 1) {
-    // There is only one unignored play; exit.
-    return 1;
-  }
```
domino14, line 230: Removing all of these early will add a little bit of time because now we have to lock and sort the plays etc first instead of exiting right away. Maybe it's not that much time.

domino14 (changes requested):
just one possible wrong loop init value

core/src/sim.c
jvc56: Yes, unless the user specified to run the sim indefinitely.

core/src/sim.c
jvc56: I thought that it would be impossible for the first play to ever be ignored and that we were only concerned with comparing every other play against the tentative winner. Let me know what I'm missing here.

core/src/sim.c
jvc56: I made a small update to ensure that the early conditions should never be true.

core/src/sim.c
domino14: you might be right here.

domino14 (approved):

PR #51, opened 2023-09-04
unhalt thread control before every sim test. fix #40

### LINKED ISSUE #40: perf_test_multithread_sim fails on MacOS
Not totally sure what's causing this, but it might be the float comparisons of zero in compare_simmed_plays are acting differently on different hardware.

domino14:


jvc56:
Resolved in another PR.

PR #58, opened 2023-09-13
some changes to make wasm more good

domino14:


core/wasmentry/api.c
```diff
@@ -1,6 +1,6 @@
 #include "../src/log.h"
 
 int main() {
-  log_set_level(LOG_INFO);
+  log_set_level(LOG_DEBUG);
```
domino14, line 4: we need to change this back to info before merging

PR #65, opened 2023-09-26
Fix Issues

jvc56:


core/src/config.c
```diff
@@ -369,69 +377,70 @@ void destroy_config(Config *config) {
...
-    strcpy(leaves, "data/lexica/french.klv2");
+    string_copy(leaves, "data/lexica/french.klv2");
   }
 
-  if (*config == NULL) {
-    *config = create_config(dist, cgp, lexicon_file, leaves, SORT_BY_EQUITY,
-                            PLAY_RECORDER_TYPE_ALL, "", "", SORT_BY_EQUITY,
-                            PLAY_RECORDER_TYPE_ALL, 0, 0, 9, 0, "", 0, 0, 0, 0,
-                            0, winpct, 100);
+  if (!*config) {
```
domino14, line 396: might it be better to be specific here? NULL is more specific (even though it is 0, yes)

core/src/gcg.c
```diff
@@ -561,15 +562,15 @@ gcg_parse_status_t parse_next_gcg_line(GCGParser *gcg_parser) {
...
       token == GCG_EXCHANGE_TOKEN) {
-    if (game_history->players[0] == NULL || game_history->players[1] == NULL) {
+    if (!game_history->players[0] || !game_history->players[1]) {
       return GCG_PARSE_STATUS_MOVE_BEFORE_PLAYER;
     }
-    if (game_history->letter_distribution == NULL) {
+    if (!game_history->letter_distribution) {
 
-      if (game_history->lexicon_name == NULL) {
+      if (!game_history->lexicon_name) {
```
domino14, line 570: same NULL v 0 question here

core/src/klv.c
```diff
@@ -93,27 +80,24 @@ float convert_little_endian_to_host(float little_endian_float) {
 
 void load_klv(KLV *klv, const char *klv_filename) {
   FILE *stream = stream_from_filename(klv_filename);
-  if (stream == NULL) {
-    perror(klv_filename);
-    exit(EXIT_FAILURE);
+  if (!stream) {
+    log_fatal("failred to open stream from filename: %s\n", klv_filename);
```
domino14, line 84: `failed`

domino14 (commented):
looks good just a q

core/src/config.c
jvc56: Leaving as is for now, might change later.

core/src/gcg.c
jvc56: Leaving as is for now, might change later.

core/src/klv.c
jvc56: Fixed in both cases.

PR #84, opened 2023-10-08
Highest equity endgame fix

olaugh:


olaugh:
@jvc56 I was wrong when I said on #macondo that the assert only failed in the inference test, it was right after that, the simming test. nwl_config uses MOVE_SORT_SCORE and i hadn't realized that when that's the case highest_possible_equity is actually highest possible score without leaves or other adjustments

jvc56:
Ah, sorry you ran into that. highest_possible_equity in this case was mislead. It is really more of a sort value than an equity in the general sense. Maybe that can be updated.

PR #88, opened 2023-10-14
Leave lookup test

olaugh:


core/test/movegen_test.c
```diff
@@ -366,6 +367,108 @@ void macondo_tests(SuperConfig *superconfig) {
...
+  Config *config = get_csw_config(superconfig);
+  Game *game = create_game(config);
+
+  char cgp[300] =
+      "ZONULE1B2APAID/1KY2RHANJA4/GAM4R2HUI2/7G6D/6FECIT3O/"
+      "6AE1TOWIES/6I7E/1EnGUARD6D/NAOI2W8/6AT7/5PYE7/5L1L7/"
+      "2COVE1L7/5X1E7/7N7 MOOORRT/BFQRTTV 340/419 0 lex CSW21;";
+  load_cgp(game, cgp);
+
+  for (int add_exchanges = false; add_exchanges <= true; add_exchanges++) {
```
jvc56, line 417: IIUC, this relies on the fact that true is implemented as 1 in stdbool, but it could in theory be any nonzero value, so I think it's preferred just keep the for loop numerical and and use i == 0 as the boolean expression to pass to generate_leaves_for_game.

jvc56 (changes requested):

core/test/movegen_test.c
olaugh: done

PR #95, opened 2023-10-23
improve get word index

andy-k:
sharing wolges's improvement here. untested.


jvc56:
LGTM. Unit test job passed with no errors.

PR #97, opened 2023-10-28
Use qsort instead of insertion sort

olaugh:
This only makes generate_moves about 0.1% faster but I'd like to add a step to movegen that splits anchors into bingo and nonbingo versions and so it makes sense to sort later (plus this will increase the number of anchors and make the sort algo matter more).


jvc56:
This is certainly an improvement over the current insert_anchor function. LGTM.

PR #107, opened 2023-11-14
Implement CGP compliant multichar handling

jvc56:


core/src/letter_distribution.c
```diff
@@ -153,47 +153,51 @@ uint8_t human_readable_letter_to_machine_letter(
...
+        // multichar exceeded max char length
+        return -1;
       }
-      // possible letter goes from index i to j. Search for it.
-      char possible_letter[MAX_LETTER_CHAR_LENGTH];
-      memory_copy(possible_letter, str + i, j - i);
-      possible_letter[j - i] = '\0';
+      current_letter[current_letter_byte_index] = current_char;
+      current_letter_byte_index++;
+      current_letter[current_letter_byte_index] = '\0';
```
andy-k, line 182: some cases i'm not seeing handled:
- `[` while already `building_multichar_letter` (not an error?)
- reaching end of `num_bytes` while still `building_multichar_letter` (not an error?)
- starting with `[]` or `]` (undefined behavior/crashes because `current_letter` is uninitialized?)
- `[]` or `]` after first `ml` while currently `!building_multichar_letter` (repeats the last single- or multichar letter?)
- out-of-bounds for `ml` because size is unknown? (already noted in comment)

writing this `\0` only after the whole series, rather than after each byte, may fix some of the above, but a different approach is to make `human_readable_letter_to_machine_letter` take a pointer and length and use something like `memcmp` (or `strncmp`?) and skip copying `current_letter` at all

core/src/letter_distribution.c
```diff
@@ -153,47 +153,51 @@ uint8_t human_readable_letter_to_machine_letter(
...
-    for (int j = i + letter_distribution->max_tile_length; j > i; j--) {
-      if (j > num_bytes) {
-        continue;
+  // Use +1 for the null terminator
+  char current_letter[MAX_LETTER_CHAR_LENGTH + 1];
+  int current_letter_byte_index = 0;
+  bool building_multichar_letter = false;
+
+  for (int i = 0; i < num_bytes; i++) {
+    char current_char = str[i];
```
andy-k, line 170: `for (char current_char; (current_char = *str); ++str) { ... }` may be faster than two passes (`string_length` is O(n))

core/src/letter_distribution.c
jvc56: Sorry, I'm a bit confused. What does out-of-bounds for `ml` mean? What does size refer to? Which comment are you referring to, your own in this review or a comment in the code?

andy-k: this comment:
https://github.com/jvc56/MAGPIE/blob/b9445b4e166ec3b50cbc2a262289b20660b829f5/core/src/letter_distribution.c#L153-L160

> it is the caller's responsibility to make this array big enough.

for example,
https://github.com/jvc56/MAGPIE/blob/b9445b4e166ec3b50cbc2a262289b20660b829f5/core/src/rack.c#L84-L98

with a long `rack_string` would overwrite `mls[MAX_BAG_SIZE]` (likely corrupting stack) before getting to `if (num_mls > MAX_BAG_SIZE) { log_fatal(...); }`, because `str_to_machine_letters` doesn't know the size it's allowed to write.

an unintuitive way to secure `set_rack_to_string` notes that _currently_ one output byte corresponds to one or more input chars, and `str_to_machine_letters` will not write to `mls[string_length(rack_string)]`:
```c
   uint8_t mls[MAX_BAG_SIZE];
   size_t len = string_length(rack_string);
   if (len > MAX_BAG_SIZE) {
     log_fatal("rack overflow: %zu\n", len); /* is %zu gcc-specific? */
   }
   int num_mls =
       str_to_machine_letters(letter_distribution, rack_string, false, mls);
```

jvc56: Ah, I see, I will work on these changes.

jvc56:
I believe the latest commit addresses all of the bugs mentioned but I could have missed something. To handle out of bounds for ml I decided to have the caller pass in the size of the ml array. I think using string_length here and doing two passes is fine since this is not on the critical path.

core/src/rack.c
```diff
@@ -86,11 +86,8 @@ int set_rack_to_string(Rack *rack, const char *rack_string,
   reset_rack(rack);
 
   uint8_t mls[MAX_BAG_SIZE];
-  int num_mls =
-      str_to_machine_letters(letter_distribution, rack_string, false, mls);
-  if (num_mls > MAX_BAG_SIZE) {
-    log_fatal("rack overflow: %d\n", num_mls);
-  }
+  int num_mls = str_to_machine_letters(letter_distribution, rack_string, false,
+                                       mls, MAX_BAG_SIZE);
```
andy-k, line 90: is it intended to no longer [`abort()`](https://github.com/jvc56/MAGPIE/blob/c52c7aa5810dc15063658c4534eef3c6bca7b1bc/core/src/log.c#L159) on invalid input?

core/src/game.c
```diff
@@ -35,10 +35,11 @@ game_variant_t get_game_variant_type_from_name(const char *variant_name) {
 cgp_parse_status_t place_letters_on_board(Game *game, const char *letters,
                                           int row_start,
                                           int *current_column_index) {
-  uint8_t *machine_letters =
-      malloc_or_die(sizeof(uint8_t) * string_length(letters));
-  int number_of_machine_letters = str_to_machine_letters(
-      game->gen->letter_distribution, letters, false, machine_letters);
+  int letters_length = string_length(letters);
```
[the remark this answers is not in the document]
jvc56: Made letters_length in game.c and mls_size in str_to_machine_letters both size_t

core/src/gcg.c
```diff
@@ -353,11 +353,11 @@ uint8_t *convert_tiles_string_to_machine_letters(
   char *tiles_string =
       get_matching_group_as_string(gcg_parser, gcg_line, group_index);
 
-  uint8_t *machine_letters =
-      malloc_or_die(sizeof(char) * (matching_group_length + 1));
+  int machine_letters_size = matching_group_length + 1;
+  uint8_t *machine_letters = malloc_or_die(sizeof(char) * machine_letters_size);
```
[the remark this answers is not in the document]
jvc56: Good point, changed.

core/src/letter_distribution.c
```diff
@@ -153,47 +156,75 @@ uint8_t human_readable_letter_to_machine_letter(
...
+      if (num_mls >= mls_size) {
+        return -1;
       }
-      // possible letter goes from index i to j. Search for it.
-      char possible_letter[MAX_LETTER_CHAR_LENGTH];
-      memory_copy(possible_letter, str + i, j - i);
-      possible_letter[j - i] = '\0';
       uint8_t ml = human_readable_letter_to_machine_letter(letter_distribution,
-                                                           possible_letter);
+                                                           current_letter);
```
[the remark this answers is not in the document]
jvc56: This is a good question, I think it's okay to leave for now, but I'll ask cesar to clarify the spec.

core/test/gcg_test.c
```diff
@@ -157,11 +157,13 @@ void assert_game_event(const GameHistory *game_history, int event_index,
     if (move_type != GAME_EVENT_PASS) {
       assert(move->tiles_played == tiles_played);
       assert(move->tiles_length == tiles_length);
+      int machine_letters_size = tiles_length + 1;
```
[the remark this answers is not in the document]
jvc56: Ah, I think this was a relic from when tiles_length == tiles_played + 1 for exchanges. I may have swapped tiles_length and tiles_played without getting rid of the +1.

core/src/letter_distribution.c
```diff
@@ -153,47 +156,75 @@ uint8_t human_readable_letter_to_machine_letter(
...
+        return -1;
+      }
+      current_letter[current_letter_byte_index] = current_char;
+      current_letter_byte_index++;
+      current_letter[current_letter_byte_index] = '\0';
+      write_machine_letter = !building_multichar_letter;
+      break;
+    }
+
+    if (write_machine_letter) {
```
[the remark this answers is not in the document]
jvc56: Removed 'write_machine_letter'

core/src/rack.c
jvc56: Yeah I don't think we should abort for overflow or any other failure. Checking if rack was set is the callers responsibility.

core/src/letter_distribution.c
```diff
@@ -153,47 +156,71 @@ uint8_t human_readable_letter_to_machine_letter(
...
-// enough.
+// enough. This function will return -1 if it encounters an invalid letter.
 // Note: This is a slow function that should not be used in any hot loops.
 int str_to_machine_letters(const LetterDistribution *letter_distribution,
                            const char *str, bool allow_played_through_marker,
-                           uint8_t *mls) {
+                           uint8_t *mls, size_t mls_size) {
 
   int num_mls = 0;
   int num_bytes = string_length(str);
```
andy-k, line 166: size_t

core/src/letter_distribution.c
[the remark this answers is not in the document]
andy-k: ok, can adjust separately if needed

andy-k (commented):
seems ok, the other changes aren't specific to this PR

core/src/letter_distribution.c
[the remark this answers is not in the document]
andy-k: conversely, do multibyte codepoints such as `ÆØÅÄÖÜ` require `[]`?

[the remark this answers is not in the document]
domino14: the way I'd thought of this, no, it doesn't require `[]`. I treat those as single "characters".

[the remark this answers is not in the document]
andy-k: it seems this PR isn't currently handling those.

jvc56:
I think this is good to merge. We can address outstanding issues in another PR.

core/src/letter_distribution.c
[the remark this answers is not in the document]
jvc56: Ah, I see. I'll need to handle the multibyte UTF8 data that isn't closed by [].

jvc56:
I believe this allows characters such as `Å` without any `[]` required. This also prevents using single characters within brackets like `[Å]`.

core/src/letter_distribution.c
```diff
@@ -150,50 +153,148 @@ uint8_t human_readable_letter_to_machine_letter(
...
+    case MULTICHAR_START_DELIMITIER:
+      if (building_multichar_letter || current_unicode_bytes_remaining > 0) {
+        return -1;
+      }
+      building_multichar_letter = true;
+      break;
+    case MULTICHAR_END_DELIMITIER:
+      // Return an error if
+      // - multichar is building
+      // - multichar has fewer than two bytes
```
andy-k, line 215: - "bytes" isn't the intent here
- does C allow //-comments?

core/src/letter_distribution.c
```diff
@@ -150,50 +153,148 @@ uint8_t human_readable_letter_to_machine_letter(
...
+      // - unicode code point is being built
+      if (!building_multichar_letter || number_of_chars_in_builder < 2 ||
```
andy-k, line 217: number_of_chars_in_builder may need a better name

core/src/letter_distribution.c
```diff
@@ -150,50 +153,148 @@ uint8_t human_readable_letter_to_machine_letter(
...
+          current_unicode_bytes_remaining > 0) {
+        return -1;
+      }
+      building_multichar_letter = false;
+      break;
+    default:
+      if (current_letter_byte_index == MAX_LETTER_CHAR_LENGTH) {
```
andy-k, line 224: MAX_LETTER_CHAR_LENGTH may need a better name since it's in bytes

core/src/letter_distribution.c
```diff
@@ -150,50 +153,148 @@ uint8_t human_readable_letter_to_machine_letter(
...
+        // Exceeded max char length
+        return -1;
+      }
+
+      int bytes_remaining =
```
andy-k, line 229: bytes_remaining may need a better name (i keep reading it as the bytes remaining for the current codepoint)

andy-k (commented):
seems ok

jvc56:
Renamed some variables to try and make a clearer distinction between bytes and letters. C99 allows // comments, which is the standard MAGPIE uses.

PR #115, opened 2023-11-27
Privatize Everything

jvc56:


jvc56:
This is a massive refactor in which almost all files are changed. The biggest changes are:

- All structs (except ones used solely for testing) are privatized by obscure pointers.
- Stateful entities and procedural code has been separated (for the most part). For all structs related to commands (inference, simming, autoplay, and movegen), results that must persist are moved to the \<command\>\_results struct in the stateful part of the code (for movegen this is just the MoveList). Transient structs needed to execute procedural code are created and destroyed entirely within the respective .c files. These structs are not visible anywhere else (though the move generator caches MoveGen structs statically since recreating them would be expensive).
- All entity functions are named "\<entity_name\>_\<function_name\>".
- All constants and enums shared across files are moved to their respective \*\_defs.h file in def
- Separate subdirectories have been created:
  - def: constants and enums related to the entity or procedure
  - ent: stateful entities which implement the obscure pointers
  - impl: procedural code which acts upon the entities
  - str: functions for printing human readable and UCGI output
  - util: various utilities
- The test config was removed. This was an unnecessary holdover from way back in the day when the config was expensive to generate. Now we have way better data structures and a fileproxy, so generating a new config for each test is no problem. It also eliminates the possibility of testing bugs due to accidental persistent changes to the config across tests, which was annoying.
- Wrote a short python script to identify circular dependencies that runs in the Github Actions.
- Removed unneeded includes with IWYU.
- Established an ordering for include statements.
- All FIXMEs are resolved.

Since some of the testing relied on the visibility of stateful structs, the affected tests were refactored (such as leave_lookup_test in move_gen_test.c) or given access to private headers. The only test requiring a private header for now is the shadow_test. Because the MoveGen struct was made private and the procedural move generator code only returns the MoveList, there was no way for the test to access the privatized anchor list. No part of the code has a valid reason to access the anchor list, but this is an important implementation detail that I think should still be tested. In general, we should only introduce more private headers if we want to whitebox test the implementation details.

I expect this to be the last major refactor for the foreseeable future. The remaining issues are related to new features, performance improvements, or are otherwise minor. MAGPIE should be fairly stable after this is merged.

PR #121, opened 2024-01-02
some more caching to generator struct in movegen

olaugh:


olaugh:
about a 11% speedup to autoplay as of [303dd96](https://github.com/jvc56/MAGPIE/pull/121/commits/303dd96cf1e59b184ad0f47802a4c083ddaaf9f6). almost all the speedup is from generate_moves but some of the kwg improvements also help play_move (netting like 0.1% across autoplay) because it uses kwg to update crosses

olaugh:
the most significant savings in this vein remaining are in recursive_gen and its callees

anything touching player->rack, gen->leave_map or gen->board is doing distant memory access

culprits are
take_letter_and_update_current_index
add_letter_and_update_current_index
get_cross_set (mostly already eliminated but still used for one-tile plays because those crosses aren't in the row cache)
score_move is the worst offender, still using
get_bonus_square
get_letter
get_cross_score
is_empty

speeding up Rack and LeaveMap would require keeping all their data local in fixed-size arrays instead of malloc'd
i briefly tried changing score_move to use the row caches but there was some problem, maybe due to board transposition. i'll look at it more carefully at somepoint. first i should get all the tests updated and passing to make sure everything is still correct before i proceed

olaugh:
The Generator -> MoveGen change already got rid of following the gen->bag and gen->player pointers which was surprisingly costly 

This removes most a lot of the gen->board and usage as well as reducing the reads of the KWG and ends up being a 5%-ish movegen speedup

olaugh:
Compared to a three-week-old version of MAGPIE and saw that things had once been much faster because of inlining. I inlined enough of the essentials (KWG, Rack, LeaveMap) to get us back to what we had.

This PR is nearly 40% faster than current main!

olaugh:
Using profiling in Instruments as a guide, I inlined functions for Anchor, Board, KLV, KWG, LeaveMap, LetterDistribution, and Rack used in move finding. 45% faster than main branch

jvc56 (approved):

Issue #122, opened 2024-01-06
library to filter the wordlist based on already-played tiles


olaugh:
inputs are a game position and optionally a previously reduced override KWG
output for now is a list of possible words

perhaps eventually it'll simply return a pointer to a new KWG without exposing the array of words but for testing and proof of concept this is a good place to start

PR #124, opened 2024-01-06
word_prune

olaugh:
this is for #122

asking for review because this is my first PR since the style changed. i would probably have more tests but i made private what seemed like it should be under the new conventions and there's no FRIEND_TEST in C. but the test for the generate function ought to cover everything, i think

some notes:

DictionaryWord (+List) is separated to be reused for the kwg maker. It is a fixed-max-size word without playthrough markers or blanking. When I do KWGs I will need to make the max length BOARD_DIM+1 instead of BOARD_DIM to allow for the gaddag separator. 

this could further reduce the dictionary if it restricted impossible hooks, and i'd like to do that relatively soon but i'd first like to get a proof of concept using the output of this to build a kwg and validating that movegen misses nothing.

i don't like the BoardRows thing and I don't think it'll be very useful when i'm calculating cross sets because location of squares on the board in both dimensions will matter a lot more.

when i handle hooks i imagine a flow like this

find max nonplaythrough size and generate words (same as now)

find playthrough anchors (mostly not whole rows) without hooks and generate their words (can dedupe here, .....D or ...D... find everything that ...D finds so ...D can be skipped)

find playthrough anchors with hooks (i'm handwaving a bit what "with hook" and "without hook" means: basically we want to be as permissive as necessary but can drop a lot of words focusing on the squares that currently have nontrivial cross sets and looking at what those cross sets can be transformed into by extensions)

i'm a bit hazy on the dependencies of the cross sets so i will have to try it out and see what changes at each step but SOMETHING LIKE:
generate horizontal cross sets
generate vertical cross sets (applying horizontal cross sets in the process)
regenerate horizontal cross sets using vertical cross sets and generate words
generate vertical words


olaugh:
Another way of thinking about hooks: we're try to answer the question of what is the *least* restrictive the cross set can ever be. It is valid and safe to assume all twenty six letters can work but it leaves us with more words in our list than we'd like. If there is a nontrivial cross set, that's good, that helps us, but we need to be careful to track how the current cross set can be relaxed to fit other letters: our cross set needs to be the union of the cross sets for all possible single letters or words surrounding the square.

jvc56 (approved):

PR #129, opened 2024-01-20
fix tests -- point to new paths

domino14:


jvc56:
This was fixed in #130

PR #144, opened 2024-02-16
Support Dynamic Board Layouts

jvc56:


jvc56:
The changes are now on top of some performance improvements in main that might mitigate slowdowns due to dynamic boards, but we'll need some perf testing to see the impact.

jvc56:
The changes still make it too slow. We will implement different board sizes by compiling with different values for BOARD_DIM.

PR #159, opened 2024-02-23
Make board transpose a parameter

jvc56:


jvc56:
I tested this to be 1% slower than main so I'm working on an alternative in #160.

jvc56:
Made irrelevant by #160

PR #160, opened 2024-02-25
Board Refactor

jvc56:


src/impl/move_gen.c
```diff
@@ -894,19 +829,10 @@ void generate_moves(Game *game, move_record_t move_record_type,
...
-    load_cross_set_cache(
-        gen, gen->current_row_index, !board_is_dir_vertical(gen->dir),
-        board_get_cross_set_index(gen->kwgs_are_shared, gen->player_index));
-    load_hz_cross_set_cache(
-        gen, gen->current_row_index,
-        board_get_cross_set_index(gen->kwgs_are_shared, gen->player_index));
+    bool anchor_transposed = anchor_get_transposed(anchor_list, i);
+    gen->row_cache =
+        board_get_const_grid(gen->board, anchor_transposed)->squares +
+        gen->current_row_index * BOARD_DIM * sizeof(Square);
```
olaugh, line 835: i know this is wip and maybe you've already fixed this locally but i think you don't want the * sizeof(Square) here. square_ptr + k already adds k*sizeof(Square) to the address

src/ent/board.h
```diff
@@ -3,157 +3,462 @@
...
+static inline void board_reset_anchor(Board *b, int row, int col, int dir) {
+  grid_reset_anchor(board_get_mutable_grid(b, b->transposed), row, col, dir);
+  grid_reset_anchor(board_get_mutable_grid(b, 1 - b->transposed), col, row,
+                    1 - dir);
+}
+
+static inline void board_reset_number_of_anchor_rows(Board *b) {
+  for (int i = 0; i < 2; i++) {
+    Grid *g = board_get_mutable_grid(b, i);
+    for (int i = 0; i < BOARD_DIM; i++) {
```
olaugh, line 353: please ignore me if you fixed this already but j instead of i for the inner loop would be more readable. (don't think it has any actually effect because it uses the i at the right scope)

jvc56: oops, didn't mean to do that, I'll fix that. I should have some compiler warnings for variable shadowing.

src/ent/board.h
```diff
@@ -182,76 +473,182 @@ static inline bool board_are_all_adjacent_squares_empty(const Board *board,
...
+  Board *board = malloc_or_die(sizeof(Board));
+  board_reset(board);
+  board_init_bonus_squares(board);
+  return board;
 }
 
-static inline bool board_get_path_is_valid(const Board *board) {
-  return board->path_is_valid;
+static inline void board_copy(Board *dst, const Board *src) {
+  memory_copy(dst, src, sizeof(Board));
```
olaugh, line 603: make memory_copy static inline

olaugh, line 603: make memory_copy static inline

jvc56: Forgot I did this in a commit I overwrote. Should be fixed now.

test/tsrc/validated_move_test.c
```diff
@@ -608,9 +622,10 @@ void test_validated_move_many() {
 }
 
 void test_validated_move() {
+  test_validated_move_distinct_kwg();
+  return;
```
olaugh, line 626: remove this early return you had for testing

jvc56: ah, thanks for catching, fixed that

olaugh (approved):

PR #166, opened 2024-03-11
Save KWG nodes on the stack before recursion

jvc56:
Improve temporal locality for KWG accesses in recursive_gen.


jvc56:
This actually made movegen slower, probably because it was dramatically increasing the stack size in recursive_gen.

PR #167, opened 2024-03-12
packed uint16_t operands getting shadow max score with avx

olaugh:
Rather shocked to see how much a speedup this gets, looks like about 3%


olaugh:
Screenshot 2024-03-11 at 5 54 44 PM

jvc56 (approved):

PR #174, opened 2024-03-17
leftx and rightx optimization for shadow

olaugh:
Reduces anchors created and anchors searched both by about 10% but speedup is very modest.
This and the other two recent "small speedup" PRs combined look to be just under a 1% speedup vs MAGPIE as of March 14.


jvc56 (approved):
LGTM! Merge this whenever you're ready.

PR #176, opened 2024-03-19
restrict unique hooks

olaugh:


olaugh:
i believe this is correct, autoplay looks good. shadow tests need to be updated

slower right now, by about 5%, mostly coming from shadow_record
set_row_restricted_hooks is slow and should be rewritten/replaced/avoided somehow
significantly less time in recursive_gen now so there's definitely upside if shadow can be tightened up

PR #177, opened 2024-03-20
Remove Anchor List Indirection

jvc56:


jvc56:
This actually makes movegen slower, likely due to suboptimal paging when shuffling around actual structs instead of pointers.

PR #179, opened 2024-03-22
incrementally sort descending effective multipliers

olaugh:
Calculating and sorting effective multiplier squares in shadow_record took up a large amount of shadow time, and did a lot of redundant work. Instead let's have MoveGen own the list and update it before recording. When the word multiplier doesn't change, it is trivial to just insert a single square's multiplier into its place. When the word multiplier changes, we do have to do more work to recalculate and reorder the list, but it's still faster than starting from scratch every time in shadow_record.

### DIFF
--- a/src/impl/autoplay.c
+++ b/src/impl/autoplay.c
@@ -20,7 +20,6 @@
 #include "move_gen.h"
 
 #include "../str/autoplay_string.h"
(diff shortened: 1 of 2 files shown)


jvc56 (approved):
LGTM! Merge when ready.

PR #181, opened 2024-03-25
fix fit-between leftx/rightx, rewrite game_gen_cross_set

olaugh:
This fixes a bug in the leftx/rightx sets which I introduced in #174. I believe they weren't set around empty squares with letters on both sides (what I call "fit-between" cross sets). The crosses were correct (uhhh, I think, more on that below) but by leaving leftx rightx at their default TRIVIAL_CROSS_SET shadow bounds were higher than they needed to be (the performance gain is near zero though) and debugging was made confusing.

I wanted to give this some relatively thorough testing so I ran a lot of games and unfortunately I see a diff with main branch where I think main branch has it wrong!!
```
   A B C D E F G H I J K L M N O   -> Player 1                 V         473
   ------------------------------     Player 2                 L         423
 1|=     A       E N     '     = | --Tracking-----------------------------------
 2|  -   R U M B A E D   Q   -   |   0
 3|    - I     A U G I T I C     |
 4|'     L       X   V   - L   ' |
 5|        J E T     O I   A     |
 6|  "       F E H   " N E P "   |
 7|    '       F O Y   G I T     |
 8|=     '       N E B E K     = |
 9|    '       '   H E N   '     |
10|  "       W       G I T   "   |
11|      W A A C       U R       |
12|'   S O R T I E D   m Y     Z |
13|    -       T R I P S E S   E |
14|  O N I O N       "   R O A D |
15|R O O D   A E M U L e S     A |
   ------------------------------
   ```
   
 Main branch appears to miss the only V play of ARVO e11 but this PR finds it.
 Agonizingly though, when I put a test case for it in main branch loading the position from a cgp and calls game_gen_all_cross_sets instead of incrementally updating the cross sets, it does find ARVO.
 
Random seed 1338 produces this game roughly 560 games in, so I will just rerun with more debug output to try to track down the issue. Based on what I know so far I think it's likeliest that this PR fixes a problem rather creating one. If so I don't know when it was introduced, but my recent shadow changes are now under some suspicion. I will want to come up with a test case before proceeding.

### DIFF
--- a/src/ent/game.c
+++ b/src/ent/game.c
@@ -247,6 +247,7 @@ void game_gen_cross_set(Game *game, int row, int col, int dir,
                                              cross_set_index,
                                              rightside_rightx_set);
   }
+  
   if (nonempty_to_left && nonempty_to_right) {
     uint64_t letter_set = 0;
     if (left_lpath_is_valid && right_lpath_is_valid) {
--- a/src/ent/game.c
+++ b/src/ent/game.c
@@ -9,7 +9,6 @@
 #include "../def/game_defs.h"
 #include "../def/letter_distribution_defs.h"
 #include "../def/players_data_defs.h"
-
 #include "bag.h"
 #include "board.h"
 #include "kwg.h"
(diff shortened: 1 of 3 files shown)


olaugh:
> This fixes a bug in the leftx/rightx sets which I introduced in #174. I believe they weren't set around empty squares with letters on both sides (what I call "fit-between" cross sets). The crosses were correct (uhhh, I think, more on that below) but by leaving leftx rightx at their default TRIVIAL_CROSS_SET shadow bounds were higher than they needed to be (the performance gain is near zero though) and debugging was made confusing.
> 
> I wanted to give this some relatively thorough testing so I ran a lot of games and unfortunately I see a diff with main branch where I think main branch has it wrong!!
> 
> ```
>    A B C D E F G H I J K L M N O   -> Player 1                 V         473
>    ------------------------------     Player 2                 L         423
>  1|=     A       E N     '     = | --Tracking-----------------------------------
>  2|  -   R U M B A E D   Q   -   |   0
>  3|    - I     A U G I T I C     |
>  4|'     L       X   V   - L   ' |
>  5|        J E T     O I   A     |
>  6|  "       F E H   " N E P "   |
>  7|    '       F O Y   G I T     |
>  8|=     '       N E B E K     = |
>  9|    '       '   H E N   '     |
> 10|  "       W       G I T   "   |
> 11|      W A A C       U R       |
> 12|'   S O R T I E D   m Y     Z |
> 13|    -       T R I P S E S   E |
> 14|  O N I O N       "   R O A D |
> 15|R O O D   A E M U L e S     A |
>    ------------------------------
> ```
> 
> Main branch appears to miss the only V play of ARVO e11 but this PR finds it. Agonizingly though, when I put a test case for it in main branch loading the position from a cgp and calls game_gen_all_cross_sets instead of incrementally updating the cross sets, it does find ARVO.
> 
> Random seed 1338 produces this game roughly 560 games in, so I will just rerun with more debug output to try to track down the issue. Based on what I know so far I think it's likeliest that this PR fixes a problem rather creating one. If so I don't know when it was introduced, but my recent shadow changes are now under some suspicion. I will want to come up with a test case before proceeding.

I ran
bin/magpie go autoplay lex CSW21 s1 equity s2 equity r1 best r2 best numplays 1 nogp rs 1338 threads 1 i 1000
on the old branches I had lying around and it looks like #174 is the cause of the ARVO bug. Older versions are fine

jvc56 (approved):
I like this version better. I think it's more readable.

PR #185, opened 2024-03-29
kwg conversion command

olaugh:
Worked on this a couple months ago. I'm not super comfortable with the config code but I want to get this in before it rots and I think this is a good first project before adding the rack table data-building code


olaugh:
Need to update the readme and see if I left any comments or debugging code I didn't mean to

olaugh:
definitely lots of unintended format changes i'll clean up

PR #187, opened 2024-03-31
prefetch kwg arcs

olaugh:
5% speedup

Screenshot 2024-03-30 at 11 29 44 PM


jvc56 (approved):
5% is incredible! LGTM

PR #189, opened 2024-04-02
test case and draft fix for unfound leaves

olaugh:


jvc56 (approved):
Forgot about this one, my bad. LGTM.

PR #200, opened 2024-04-14
fixes for 8+ racks

olaugh:
RACK_SIZE option in Makefile
peg array size
move->tiles and gen->strip sizes need to be max(RACK_SIZE, BOARD_DIM)

Works with up to 28 on my M2 mbp with 8GB ram but be aware that LeaveMap.values is sizeof(double)*2^RACK_SIZE (2GB for 28 tiles) and there's one of those per thread.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,11 @@ ifndef BOARD_DIM
 BOARD_DIM = 15
 endif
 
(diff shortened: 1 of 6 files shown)


jvc56 (approved):
Badass. LGTM.

PR #201, opened 2024-04-14
option for printing color boards

olaugh:
sending this out as an option, lmk what you think. tiles in bold does improve legibility for me, colors for bonus squares might be a bit busy but i've been wanting to try it out for a while

make clean; make magpie PRINT_COLOR_BOARDS=true


olaugh:
Screenshot 2024-04-14 at 2 57 39 PM

jvc56:
This LGTM. I'd definitely use it.

olaugh:
> This LGTM. I'd definitely use it.

thanks! but i'm going to redo it to get rid of the #define. that only makes sense for things like BOARD_DIM and RACK_SIZE where we can't get optimal performance out of the same binary. i was mainly trying to avoid putting it as a config parameter, but i don't need to do that either. i'll just make it string_builder_add_color_game and use that when i want it. i may make this even fancier too

olaugh:
Screenshot 2024-04-17 at 11 29 10 PM

olaugh:
code cleanup coming but this is about all i'm going to do for this PR
Screenshot 2024-04-21 at 4 31 11 PM

olaugh:
obsolete, https://github.com/jvc56/MAGPIE/pull/400 is the new version

PR #204, opened 2024-05-06
KWG Cache

jvc56:
Not sure if this is faster or slower than main yet, so I'll need to do some performance testing.


jvc56:
Laughably slow.

PR #220, opened 2024-06-17
Functional config load

jvc56:
The primary motivation for this major refactor was to simplify the addition of new commands and persistent configuration variables to MAGPIE. These changes include but are not necessarily limited to:

 - Separate the command arguments from the persistant config variables.
   - This separation is mostly subjective, based on what I think would be unique for a given command and what would be tedious to reenter every time. For example, all of the arguments to the convert command are command arguments while the equity margin for inference is now a persistent config variable. This means that config fields that were just used as command arguments and 'passed through' the Config no longer pollute the Config struct.
   - Command arguments are now stored in the pargs field while persistent variable are given their own field in the Config.
 - Use function pointers to execute commands and retrieve statuses to remove tedious and repetitive code. To implement this:
   - The ExecState struct was deleted and all of it's fields were added to the Config as persistent variables.
   - The Config module now imports the various impl modules instead of the other way around, since the config is now executing these functions.
   - To prevent cyclic dependencies, argument helper structs were created for various modules, such as GameArgs, SimArgs, etc.
 - Options now require a single '-' char before the option name, for example: 'setoptions -r1 best -r2 best -it 100 -numplays 1 -threads 4'
 - All command and option names can now be the minimum length prefix string that disambiguates the command or option. For example, the above command can be executed as: 'set -r1 best -r2 b -it 100 -nump 1 -thr 4'
 - The stop condition can now take on any decimal value in [0, 100] instead of the three fixed values 95, 98, and 99. To not use a stopping condition, use 'none'. Valid examples of this option include '-scond 96.453' or '-scond none'.


README.md
```diff
@@ -275,39 +259,32 @@ Specifies the input file for the MAGPIE execution. This can be a regular file or
...
- ./bin/magpie console
-magpie>go sim static plies 1 threads 1 numplays 15 cgp C14/O2TOY9/mIRADOR8/F4DAB2PUGH1/I5GOOEY3V/T4XI2MALTHA/14N/6GUM3OWN/7PEW2DOE/9EF1DOR/2KUNA1J1BEVELS/3TURRETs2S2/7A4T2/7N7/7S7 EEEIILZ/ 336/298 0 lex NWL20;
+./bin/magpie sim -threads 7 -numplays 5 -scond 99 -it 10000 -pfreq 100 -cfreq 500 -plies 5
 ```
 
 Run an inference:
 
 ```
- ./bin/magpie go infer rack HUJA pindex 0 score 20 exch 0 numplays 20 info 1000000 threads 4 lex OSPS44
+ ./bin/magpie infer 0 HUJA 20 -numplays 20 -pfreq 1000000 -threads 4 -lex OSPS44
```
olaugh, line 281: player indices are now 1 and 2, not 0 and 1

olaugh (approved):

README.md
jvc56: Thanks for catching! Just fixed.

PR #248, opened 2024-09-07
fix clang compilation errors

olaugh:
These are preventing dev build for me

first this, same problem as in https://github.com/jvc56/MAGPIE/pull/239/files
```
src/impl/convert.c:119:3: error: variable 'filepath_type' is used uninitialized whenever switch default is taken [-Werror,-Wsometimes-uninitialized]
  default:
  ^~~~~~~
src/impl/convert.c:123:10: note: uninitialized use occurs here
  return filepath_type;
         ^~~~~~~~~~~~~
src/impl/convert.c:96:3: note: variable 'filepath_type' is declared here
  data_filepath_t filepath_type;
  ^
src/impl/convert.c:151:3: error: variable 'filepath_type' is used uninitialized whenever switch default is taken [-Werror,-Wsometimes-uninitialized]
  default:
  ^~~~~~~
src/impl/convert.c:155:10: note: uninitialized use occurs here
  return filepath_type;
         ^~~~~~~~~~~~~
src/impl/convert.c:128:3: note: variable 'filepath_type' is declared here
  data_filepath_t filepath_type;
  ^
2 errors generated.
```

and then this–hard to believe given how natural it is to write this code–is actually totally forbidden in C23. we have "-Werror,-Wstrict-prototypes" in our Makefile so I don't know why it's building in CI.

```
src/ent/autoplay_results.c:467:41: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
AutoplayResults *autoplay_results_create() {
                                        ^
                                         void
1 error generated.
make: *** [obj/src/ent/autoplay_results.o] Error 1
```


jvc56:
Weird. I also would have thought the "-Werror,-Wstrict-prototypes"  would cause an error. Anyway, thanks for fixing.

jvc56 (approved):

PR #250, opened 2024-09-08
128 bit rack storage, needed as keys for maps in  'heuri' algo

olaugh:


jvc56 (approved):
LGTM. Just so you know, in addition to this bit field rack you implemented in this PR, we also have a run length encoded rack at

https://github.com/jvc56/MAGPIE/blob/main/src/ent/encoded_rack.h

Which obviously wouldn't work in the Heuri algo since the bit location of each machine letter needs to be fixed, but it's there if you ever need it for something. It's a bit more generic in that it uses an array of uint64_t instead of just high and low, so if encoded rack ever becomes a bottleneck somewhere we can use the intrinsic 128 implementation you've done here to make it more performant. Thanks for the PR!

olaugh:
> LGTM. Just so you know, in addition to this bit field rack you implemented in this PR, we also have a variable-length encoded rack at
> 
> https://github.com/jvc56/MAGPIE/blob/main/src/ent/encoded_rack.h
> 
> Which obviously wouldn't work in the Heuri algo since the bit location of each machine letter needs to be fixed, but it's there if you ever need it for something. It's a bit more generic in that it uses an array of uint64_t instead of just high and low, so if encoded rack ever becomes a bottleneck somewhere we can use the intrinsic 128 implementation you've done here to make it more performant. Thanks for the PR! You can merge this at your convenience.

Interesting, thanks! Sure, let's look at unifying these or sharing techniques later. I saw you merging frequently and thought I should probably start doing the same. My bingo gen code was in these big multi-month PRs that ended up rotting and I want to make sure this doesn't end up like this, and it seemed like a good time to do it because we'll probably have very few conflicts while you're working on leaves.

PR #258, opened 2024-09-19
wordmap builder

olaugh:


olaugh:
Lots of work yet to do but I wanted to make sure this existed somewhere besides my laptop.

src/ent/bit_rack.h
```diff
@@ -1,12 +1,19 @@
 #ifndef BIT_RACK_H
 #define BIT_RACK_H
 
+#if defined(__APPLE__)
+#include "../../compat/endian.h"
+#else
+#include <endian.h>
+#endif
+
```
jvc56, line 9: We now handle this in compat/endian_conv.h instead of checking the OS in a bunch of different files. If new endian function are needed you can add them to compat/endian_conv.h

src/ent/bit_rack.h
```diff
@@ -50,16 +63,49 @@ bool bit_rack_is_compatible_with_ld(const LetterDistribution *ld) {
...
+  return sizeof(BitRack) == 16;
+}
+
+static inline bool bit_rack_is_compatible_with_endianness(void) {
+#if defined(__APPLE__)
+  return OSHostByteOrder() == OSLittleEndian;
+#else
+  return __BYTE_ORDER == __LITTLE_ENDIAN;
+#endif
+}
```
jvc56, line 76: Same as above. Prefer to handle in compat/endian_conv.h

src/ent/bit_rack.h
```diff
@@ -156,20 +206,208 @@ void bit_rack_div_mod(const BitRack *bit_rack, uint32_t divisor,
...
+    bit_rack->low -= (uint64_t)1 << shift;
+  } else {
+    bit_rack->high -= (uint64_t)1 << (shift - 64);
+  }
+#endif
+}
+
+static inline void bit_rack_write_12_bytes(const BitRack *bit_rack,
+                                           uint8_t bytes[12]) {
+  // assumes little-endian
```
jvc56, line 343: Nit (only change if convenient): if possible try to remain endian-ambiguous and use compat/endian_conv.h if that helps.

src/ent/bit_rack.h
```diff
@@ -156,20 +206,208 @@ void bit_rack_div_mod(const BitRack *bit_rack, uint32_t divisor,
...
+  memory_copy(bytes, ((uint8_t *)bit_rack), 12);
+#else
+  memory_copy(bytes, ((uint8_t *)&bit_rack->low), 8);
+  memory_copy(bytes + 8, ((uint8_t *)&bit_rack->high), 4);
+#endif
+}
+
+static inline BitRack bit_rack_read_12_bytes(const uint8_t bytes[12]) {
+  BitRack bit_rack = bit_rack_create_empty();
+  // assumes little-endian
```
jvc56, line 354: Same as above.

test/tsrc/test.c
```diff
@@ -41,10 +41,15 @@
 #include "string_util_test.h"
 #include "validated_move_test.h"
 #include "wasm_api_test.h"
+#include "wmp_test.h"
+#include "wmp_maker_test.h"
 #include "word_prune_test.h"
 #include "word_test.h"
 
 void run_all(void) {
+  test_wmp();
```
jvc56, line 50: Make sure we run the full test before merging.

jvc56 (changes requested):

src/ent/bit_rack.h
olaugh: i believe the last common big endian architecture for personal computing was powerpc (i had one)
but it's good to be hygienic about this stuff, and actually having thought more about it, it wouldn't be hard to be compatible (but it would take some effort to emulate a big endian machine)
i can fix this up

test/tsrc/test.c
olaugh: oops my bad, fixed. i still haven't figured out how to detect leaks locally so i've been relying heavily on github ci, and it was taking too long to run the whole set of sets, so that's what i was up to

src/ent/bit_rack.h
olaugh: Removed this, only including endian_conv.h

src/ent/bit_rack.h
olaugh: Removed this, this check no longer exists, compatible with either endianness (though big endian is untested)

src/ent/bit_rack.h
olaugh: same kind of fix

src/ent/bit_rack.h
olaugh: I have done the endian swap using existing endian_conv.h functions for 64 and 32 but I didn't add these 96 bit swaps there, they just seem overly specific to this use case, especially since it depends on whether BitRack is  __int128 or a struct

olaugh:
Thank you and PTAL!

jvc56 (approved):

PR #269, opened 2024-11-10
use wordmap existence checks in shadow

olaugh:


src/impl/wmp_move_gen.h
```diff
@@ -117,7 +117,7 @@ wmp_move_gen_check_playthrough_full_rack_existence(WMPMoveGen *wmp_move_gen) {
   // position, but only one combination is possible, and the leave is empty.
   const int size = wmp_move_gen->full_rack_size;
   const int idx = subracks_get_combination_offset(size);
-  SubrackInfo *playthrough_info = &wmp_move_gen->nonplaythrough_infos[idx];
+  SubrackInfo *playthrough_info = &wmp_move_gen->playthrough_infos[idx];
```
olaugh, line 120: I was initially surprised this bug didn't break anything in autoplay but I ran 100k game pairs and it didn't make a difference. In this branch we've already done everything we're going to do with the nonplaythrough SubrackInfos, we looked up whether they had words before shadow started.

However in my work using the info to find plays I was very confused to see letters inserted into the nonplaythrough BitRacks which of course shouldn't ever be bigger than RACK_SIZE.

PR #270, opened 2024-11-11
shrink wmp some by storing double blanks with bitvectors instead of l…

olaugh:
shrink wmp file some by storing double blanks with bitvectors instead of lists of pairs


olaugh:
back to draft, i see some cleanup i want to do in wmp_maker

olaugh:
I can tell I didn't do the best job cleaning up WMPEntry because it's not easy to resize it. There's some functions still using pointer arithmetic instead of the union fields in the struct, but even having fixed that locally it's still got problems. Workin' on it

olaugh:
> I can tell I didn't do the best job cleaning up WMPEntry because it's not easy to resize it. There's some functions still using pointer arithmetic instead of the union fields in the struct, but even having fixed that locally it's still got problems. Workin' on it

Fixed that and tested with WMP_INLINE_VALUE_BYTES set to various multiples of 4: 12, 16, 20, 32. Leaving the value at 16

olaugh:
CSW21.wmp before: 162502154
CSW21.wmp now: 149080312

olaugh:
> CSW21.wmp before: 162502154 bytes CSW21.wmp now: 149080312
> 
> With WMP_INLINE_VALUE_BYTES set to 12, would be 130989842

fwiw the new CSW21.wmp gzips to under 50MB, fairly reasonable size if we have to send it over network

PR #282, opened 2025-01-06
Endgame

domino14:
Implement negamax-based exhaustive endgame engine. Optimizations include:
- alpha-beta pruning
- iterative deepening
- negascout
- small move (128 bits describes every move) recorder for movegen
- arena-like (really stack-like) allocations for moves

Optimizations not yet included, but in future PRs:
- transposition table
- LazySMP or better multithreading
- wordprune


jvc56 (approved):
Just left one note about the cross platform stuff. Also, it would be good to clean up some of the commented prints we don't need anymore, unless we'll need them for future debugging. Other than that it looks good to go.

src/impl/endgame.c
```diff
@@ -0,0 +1,444 @@
+#include <pthread.h>
+
+#include "../../src/ent/move.h"
```
olaugh, line 3: #include "../ent/move.h"

olaugh (approved):

domino14:
> Just left one note about the cross platform stuff. Also, it would be good to clean up some of the commented prints we don't need anymore, unless we'll need them for future debugging. Other than that it looks good to go.

Yep, let's leave them for a little bit of time for future debugging, then we can remove i think after adding all the other stuff from Macondo.

PR #285, opened 2025-01-12
Use Equity in movegen + elsewhere

olaugh:


src/ent/game.c
```diff
@@ -236,7 +239,7 @@ static inline void game_gen_alpha_cross_set(Game *game, int row, int col,
       board_get_word_edge(board, row, col - 1, WORD_DIRECTION_LEFT);
   const int right_col =
       board_get_word_edge(board, row, col + 1, WORD_DIRECTION_RIGHT);
-  int score = 0;
+  Equity score = 0;
```
jvc56, line 242: Not sure if we want to use EQUITY_ZERO_VALUE here. I think it's fine either way, so I'll leave it up to you.

src/ent/static_eval.h
```diff
@@ -176,10 +175,10 @@ static inline int static_eval_get_move_score(const LetterDistribution *ld,
     col_start = tmp_start;
   }
 
-  int ls;
-  int main_word_score = 0;
-  int cross_scores = 0;
-  int word_bingo_bonus = 0;
+  Equity ls;
+  Equity main_word_score = 0;
```
jvc56, line 179: Not sure what the difference between setting to 0 and setting to EQUITY_ZERO_VALUE is here, but it seems intentional, so just let me know the logic behind it.

src/ent/move.h
```diff
@@ -318,10 +321,6 @@ static inline void small_move_destroy(SmallMove *move) {
   free(move);
 }
 
-static inline bool within_epsilon_for_equity(double a, double b) {
```
jvc56, line 321: Very glad to get rid of this atrocity.

jvc56 (approved):

PR #286, opened 2025-01-14
trim whitespace csv fields in csv2klv conversion

olaugh:
I had trouble converting the csv i got from Gilles because of dos line endings (CR+LF). This seems good to have. I tested it but handling it in convert_test was annoying because CSW21_small.csv gets overwritten so I punted


olaugh:
this removes {' ', '\t', '\n', '\v', '\f', '\r'}

jvc56 (approved):

PR #293, opened 2025-02-13
use fixed-size arrays in board/move/rack/ld

olaugh:
ints shrunk and reordered for packing, pointers to malloc'ed blocks replaced by fixed-length arrays within the structs


olaugh:
3% speedup

PR #367, opened 2025-07-30
Load gcg testing

truthmane:


src/impl/load.h
```diff
@@ -0,0 +1,24 @@
...
+  void *config;
+} DownloadGCGOptions;
+
+// helper functions for .gcg sources
+char *get_xt_gcg_string(const char *identifier, ErrorStack *error_stack);
+char *get_woogles_gcg_string(const char *identifier, ErrorStack *error_stack);
+char *get_local_gcg_string(const char *identifier, ErrorStack *error_stack);
+
+// main dispatcher function
+char *get_gcg_string(const DownloadGCGOptions *options, ErrorStack *error_stack);
```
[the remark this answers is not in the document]
truthmane: used to c++ professors telling me that i have to declare every single function in the header lol

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+  char *game_id_str = NULL;
+
+  // Check if this is a Cross-tables URL
+  if (strstr(identifier, "cross-tables.com") != NULL) {
+      // Extract game ID from URL parameter u=XXXXX using regex
+      regex_t regex;
+      regmatch_t matches[2];
+      
+      int regex_status = regcomp(&regex, "u=([0-9]+)", REG_EXTENDED);
+      if (regex_status != 0) {
```
jvc56, line 25: since the regex is constant failing compilation would be pretty bad, so we should do log_fatal instead I think.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+          int start = matches[1].rm_so;
+          int end = matches[1].rm_eo;
+          int id_len = end - start;
+          
+          game_id_str = malloc_or_die(id_len + 1);
+          strncpy(game_id_str, identifier + start, id_len);
+          game_id_str[id_len] = '\0';
+      } else {
+          regfree(&regex);
+          error_stack_push(error_stack, ERROR_STATUS_CONFIG_LOAD_MISSING_ARG,
```
jvc56, line 42: Would be good to make a new enum for this that indicates the xt URL was missing the game id or malformed.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+      // Check if it's a numeric game id
+      if (!is_all_digits_or_empty(identifier)){
+        return NULL;
+      }
+      game_id_str = string_duplicate(identifier);
+  }
+
+  char *gcg_content = NULL;
+
+  if (game_id_str) {
```
jvc56, line 58: You can remove this if check since it's guaranteed that game_id_str is nonnull at this point.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+      
+      free(curl_cmd);
+      free(game_id_str);
+  }
+
+  if (!gcg_content || is_string_empty_or_whitespace(gcg_content) || 
+      strstr(gcg_content, "#player1") == NULL) {
+      free(gcg_content);
+      // Only push error if this was actually a cross-tables URL
+      if (strstr(identifier, "cross-tables.com") != NULL) {
```
jvc56, line 79: If the GCG content is screwed up we should probably return an error regardless of the identifier type. Some new error enum like "bad xt gcg" and a message saying the id, so that the error makes sense for both URL and game id input types.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+                return NULL;
+            }
+        }
+        // It's alphanumeric and not all digits, assume it's a woogles game ID
+        game_id_str = string_duplicate(identifier);
+    }
+    
+    char *gcg_content = NULL;
+
+    if (game_id_str) {
```
jvc56, line 143: You can remove this if check for the same reason as above.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+        // Use woogles API to get GCG content
+        char *curl_cmd = get_formatted_string("curl -s -H 'Content-Type: application/json' "
+                                              "https://woogles.io/api/game_service.GameMetadataService/GetGCG "
+                                              "-d '{\"game_id\":\"%s\"}'", game_id_str);
+        
+        char *response = get_process_output(curl_cmd);
+        free(curl_cmd);
+        free(game_id_str);
+        
+        if (response) {
```
jvc56, line 153: If the response is NULL we should just return an error which avoids the nesting.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+            // Parse JSON response to extract GCG content
+            const char *gcg_start = strstr(response, "\"gcg\":\"");
```
jvc56, line 155: Same here, if gcg_start is NULL, return with an error to avoid the nesting.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+            if (gcg_start) {
+                gcg_start += 7; // Skip "\"gcg\":\""
```
jvc56, line 157: Instead of "7" prefer to assign "\"gcg\":\"" to use a define statement at the top of the file to use above like  strstr(response, WOOGLES_GCG_START); and use gcg_start += string_length(WOOGLES_GCG_START);

That way if you need to change the string, you don't have to remember to edit two places.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+                while (*gcg_end && *gcg_end != '"') {
+                    gcg_end++;
+                }
+                int gcg_len = gcg_end - gcg_start;
+                char *raw_gcg = malloc_or_die(gcg_len + 1);
+                strncpy(raw_gcg, gcg_start, gcg_len);
+                raw_gcg[gcg_len] = '\0';
+                
+                // Unescape JSON string using utility function
+                gcg_content = json_unescape_string(raw_gcg, error_stack);
```
jvc56, line 168: Check if the error stack is nonempty here and return if it is.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+                free(raw_gcg);
+                
+                if (!gcg_content) {
```
jvc56, line 171: After adding the error stack check above, checking the gcg_content for NULL is unnecessary.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+                    free(response);
+                    return NULL;
+                }
+            }
+            
+            free(response);
+        }
+    }
+    
+    if (!gcg_content || is_string_empty_or_whitespace(gcg_content) ||
```
jvc56, line 181: Since this is repeated from above, you can make a function called is_valid_gcg_content or something similar that can be called in both places.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+        strstr(gcg_content, "#player1") == NULL) {
+        free(gcg_content);
+        // Only push error if this was actually a woogles URL
+        if (strstr(identifier, "woogles.io") != NULL) {
```
jvc56, line 185: Same as above, we should return an error for bad gcg_content for any identifier.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+}
+
+char *get_local_gcg_string(const char *identifier, ErrorStack *error_stack) {
+    
+    // Try to open as local file
+    FILE *file = fopen(identifier, "r");
+    if (!file) {
+        return NULL;  // Not a valid local file - not our job
+    }
+    fclose(file);
```
jvc56, line 203: Instead of opening and closing the file, there's probably some more concise way in C to just check if the file exists and is readable.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+    
+    // Read file content directly
+    char *gcg_content = get_string_from_file(identifier, error_stack);
```
jvc56, line 206: Check the error stack here and return if not empty.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+
+    if (!gcg_content || is_string_empty_or_whitespace(gcg_content) ||
```
jvc56, line 208: Use the function described above for this.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+        strstr(gcg_content, "#player1") == NULL) {
+        free(gcg_content);
+        error_stack_push(error_stack, ERROR_STATUS_FILEPATH_FILE_NOT_FOUND,
```
jvc56, line 211: I think a new error enum here is appropriate, something like "bad local file gcg content" or something.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+        return NULL;
+    }
+
+    return gcg_content;
+}
+
+char *get_gcg_string(const DownloadGCGOptions *options, ErrorStack *error_stack) {
+  
+    if (!options || !options->source_identifier) {
+        error_stack_push(error_stack, ERROR_STATUS_CONFIG_LOAD_MISSING_ARG,
```
jvc56, line 222: You can make this a log_fatal since it should eventually be impossible for the user to trigger this due to previous checks.

src/impl/load.c
```diff
@@ -0,0 +1,283 @@
...
+                         string_duplicate("Invalid options or missing source_identifier"));
+        return NULL;
+    }
+    const char *identifier = options->source_identifier;
+    
+    // Try Cross-tables first 
+    char *gcg_string = get_xt_gcg_string(identifier, error_stack);
+    if (!error_stack_is_empty(error_stack)) {
+        return NULL;  // Fatal error occurred
```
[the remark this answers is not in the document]
jvc56: You can remove the // Fatal error occurred comment since the error isn't fatal. Same for the ones below.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+    // If we get here, nothing worked
+    error_stack_push(error_stack, ERROR_STATUS_FILEPATH_FILE_NOT_FOUND,
+                     get_formatted_string("Could not load GCG from any source: %s", 
+                                          options->source_identifier));
+    return NULL;
+}
+
+void download_gcg(const DownloadGCGOptions *options, GameHistory *game_history, ErrorStack *error_stack) {
+    if (!options || !game_history) {
+        error_stack_push(error_stack, ERROR_STATUS_CONFIG_LOAD_MISSING_ARG,
```
jvc56, line 264: Make this a log_fatal since this shouldn't be possible.

src/impl/load.c
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    // Validate config is provided for parsing
+    Config *config_to_use = options->config;
+    if (!config_to_use) {
+        error_stack_push(error_stack, ERROR_STATUS_CONFIG_LOAD_MISSING_ARG,
+                         string_duplicate("GCG content downloaded successfully but parsing requires a valid config"));
+        free(gcg_content);
+        return;
+    }
```
jvc56, line 282: I think it's fine to delete this and assume the Config is always in the options.

src/util/string_util.c
```diff
@@ -800,3 +800,37 @@ double string_to_double(const char *str, ErrorStack *error_stack) {
...
 }
+
+char *json_unescape_string(const char *json_string, ErrorStack *error_stack) {
+  if (!json_string) {
+    error_stack_push(error_stack, ERROR_STATUS_CONFIG_LOAD_MISSING_ARG,
+                     string_duplicate("json_string cannot be null"));
+    return NULL;
+  }
+  
+  int input_len = strlen(json_string);
```
jvc56, line 811: You can remove the error check and just use string_length(json_string), which will log_fatal on NULL, and so I guess you can remove error_stack from the signature at that point.

test/load_test.c
```diff
@@ -0,0 +1,294 @@
...
+    printf("Skipping cross-tables integration test (set MAGPIE_INTEGRATION_TESTS=1 to enable)\n");
+    return;
+  }
+
+  GameHistory *game_history = game_history_create();
+  ErrorStack *error_stack = error_stack_create();
+
+  // Test with a known working cross-tables game URL
+  // Using a classic game that should be stable on cross-tables
+  Config *config = config_create_default_with_data_paths(error_stack, "data");
```
jvc56, line 27: Prefer to use config_create_or_die for all tests, that way you don't need to check for creation success.

test/load_test.c
```diff
@@ -0,0 +1,294 @@
...
+  } else {
+    // Basic validation that we got a valid game
+    assert(game_history_get_number_of_events(game_history) > 0);
+    
+    // Check that we have players
+    assert(game_history_both_players_are_set(game_history));
+    
+    printf("Cross-tables integration test passed: loaded %d game events\n",
+           game_history_get_number_of_events(game_history));
+  }
```
jvc56, line 60: This section (and similar sections below) is a good candidates for refactoring out into a new test helper function, something like "validate_download_gcg" which takes just the options as an argument and asserts that:

 - error stack is empty
 - game history has some events
 - both player names are set

You can create and destroy the game history and error stack in this new function so you don't have to do it every time for each test.

jvc56: You could also make this function handle the error cases as well by passing in the expected error stack error enum and asserting that as well.

jvc56:
Last few things before we merge:

 - Use config_create_or_die instead of other config creation function in the tests
 - Update validate_download_gcg to have the following signature:

```
void validate_download_gcg(const char *source_identifier, error_code_t expected_error)
```
and have it create it's own config and options to reduce repeated code in the tests.
 - I forgot about this before (that's my bad), but we should add a "download from URL" option that just downloads a GCG from a given URL. This should be after the woogles download but before the local download. It will be very similar to the current xt download except that you don't even have to construct the URL, you just try downloading using the given source identifier if it looks like a URL.

src/impl/load.c
```diff
@@ -20,6 +20,29 @@ char *get_xt_gcg_string(const char *identifier, ErrorStack *error_stack) {
 
   // Check if this is a Cross-tables URL
   if (strstr(identifier, "cross-tables.com") != NULL) {
+      // Check if it's a direct GCG URL first
+      if (strstr(identifier, "cross-tables.com/annotated/selfgcg") != NULL) {
```
jvc56, line 24: This one's my bad, I should have clarified that I meant a URL from any source, not necessarily cross-tables, so we should probably move this to it's own "get_url_gcg_string" function or something.

test/load_test.c
```diff
@@ -32,168 +38,86 @@ void validate_download_gcg(const DownloadGCGOptions *options, error_code_t expec
...
+  test_xt_url_conversion();
+  test_xt_game_id();
+  test_xt_direct_url();
   printf("Cross-tables tests completed.\n\n");
   
   printf("=== Woogles Tests ===\n");
-  test_woogles_url_recognition();
-  test_woogles_integration();
+  test_woogles_url_conversion();
+  test_woogles_game_id();
```
jvc56, line 120: We can simplify even further by just directly calling validate_download_gcg from here and inline the source identifier like:

```
validate_download_gcg("/tmp/nonexistent_file.gcg", ERROR_STATUS_FILEPATH_FILE_NOT_FOUND);
```

for each function call.

jvc56 (commented):
If regcomp fails for any of the woogles, xt, or URL functions, call "log_fatal" instead of returning NULL since that shouldn't be possible.

jvc56:
LGTM! Merge this in at your convenience.

PR #392, opened 2025-09-18
Stack Alloc MoveGen

jvc56:


jvc56:
Both stack allocated and static global MoveGen structs were slower than the existing heap allocated MoveGen structs.

PR #396, opened 2025-09-21
Implement Information-Directed Selection for the Top Two BAI algorithm

jvc56:


jvc56:
This PR introduces an Information-Directed Selection technique for deciding whether to sample the leader or the challenger for the Top Two algorithm. Details in https://arxiv.org/pdf/2205.12086.

PR #402, opened 2025-10-06
Per player sim params and simmed autoplay

olaugh:


src/impl/autoplay.c
```diff
@@ -320,6 +327,153 @@ void game_runner_start(const AutoplayWorker *autoplay_worker,
...
+
+  // Limit the number of plays to sim
+  if (num_moves > sim_num_plays) {
+    // Keep only the top sim_num_plays moves by truncating the count
+    // (moves are already sorted by equity from generate_moves)
+    move_list->count = sim_num_plays;
+  }
+
+  // Create sim results for this simulation
+  SimResults *sim_results = sim_results_create();
```
jvc56, line 382: We should try to create these at the beginning on a per-thread basis instead of every time during movegen.

PR #420, opened 2025-12-03
Infer cutoff test fix

olaugh:


olaugh:
This is just for reference while I hand-code something cleaner

PR #423, opened 2025-12-11
Heatmaps

jvc56:


jvc56:
I don't like how this implementation is turning out. I need to do some sim results refactoring to prepare for these changes. This branch is now just to save my current progress.

PR #431, opened 2026-01-13
Prealloc sim and infer

jvc56:
Creates Sim Context and Inference Context structs which can be used across multiple sim and inference calls to avoid repeated allocations for the eventual autoplay sim feature.


jvc56:
The Copilot review actually caught a bug which I'll have to work on fixing once I get back from NOLA. I should be able to get this in before next weekend.

PR #471, opened 2026-02-14
Add root stuck-tile detection and per-thread diversity

olaugh:
## Summary
- Change `stuck_tile_fraction_from_bv` from tile-count-weighted to score-weighted fraction (high-value stuck tiles matter more)
- Compute `initial_opp_stuck_frac` at solve start via movegen, so root move ordering uses build-chain heuristics from the first iteration
- Even-indexed threads use stuck-tile-aware root ordering, odd threads use plain score ordering for search diversity
- Add static endgame equity estimate for non-stuck path: `score - face_value_of_tiles_played` instead of raw score

## Test plan
- [x] `make clean && make magpie_test BUILD=release` compiles with no warnings
- [x] `./bin/magpie_test endgame` passes (exit 0)

### DIFF
--- a/src/impl/endgame.c
+++ b/src/impl/endgame.c
@@ -93,28 +93,29 @@ static inline uint64_t shadow_zobrist_add_move(const Zobrist *z,
   return shadow_key;
 }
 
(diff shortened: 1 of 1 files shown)


olaugh:
Folded into #470

PR #472, opened 2026-02-14
Update test callbacks and add eldar_v to test suite

olaugh:
## Summary
- Silence ranked moves in `print_pv_callback` (void the params)
- Add `print_pv_and_ranked_callback` for tests that want ranked output
- Change `test_eldar_v_stick` from 9-ply to 5-ply
- Add `eldar_v` to always-run unit test table (runs with `./run u`, separate from `test_endgame`)
- Update `test_14domino` to use ranked callback with `num_top_moves = 10`
- Rename on-demand table key from `"eldar"` to `"eldar_v"`

## Test plan
- [x] `make clean && make magpie_test BUILD=release` compiles with no warnings
- [x] `./bin/magpie_test endgame` passes (exit 0)
- [x] `./bin/magpie_test eldar_v` passes (exit 0)

### DIFF
--- a/test/endgame_test.c
+++ b/test/endgame_test.c
@@ -87,17 +87,26 @@ static void print_pv_callback(int depth, int32_t value, const PVLine *pv_line,
   printf("%s\n", string_builder_peek(sb));
   string_builder_destroy(sb);
 
-  // Print ranked root moves with full PV lines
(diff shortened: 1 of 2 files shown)


olaugh:
Folded into #470

PR #489, opened 2026-03-04
Autoplay Sims

jvc56:


jvc56:
We may have to accept the additional complexity since all of my attempts to move the MoveGen cache to the Config have yielded 1-2% performance regressions for static autoplay:

 - #490: MoveGenCache uses a dynamically allocated array of MoveGen structs.
 - #491: MoveGenCache uses a dynamically allocated, memory aligned array of MoveGen structs.
 - #494: MoveGenCache uses a dynamically allocated array of MoveGen pointers.
 - #493: MoveGenCache uses a fixed size array of MoveGen pointers.
 - #492: MoveGenCache uses a dynamically allocated array of MoveGen structs and also uses custom *best_equity_only functions to avoid stack allocating the MoveGenArgs struct.
 
@olaugh Unless you have any ideas that might be quick, I'm ready to merge this in now and deal with the global MoveGens later. A 1-2% performance regression is unacceptable IMO.

jvc56:
I confirmed there is no regression for static autoplay.

PR #511, opened 2026-04-04
Fix crash when adding note after navigating to start of game

### LINKED ISSUE #510: Adding note triggers crash
Repro:
* Create a new game with `new` command
* Set rack and commit a move
* Return to first move and add a note using `note` command

Yields this error:
```
magpie(53811,0x16b287000) malloc: *** error for object 0x7e6b00000000: pointer being freed was not allocated
magpie(53811,0x16b287000) malloc: *** set a breakpoint in malloc_error_break to debug
```

olaugh:
The note command checked game_history_get_num_events (total events) instead
of game_history_get_num_played_events (current position). When a user
navigated back to the start with "goto start" or "previous", num_played_events
was 0 but num_events was nonzero, causing an index of -1 access in
game_history_set_note_for_most_recent_event.

Fixes #510

https://claude.ai/code/session_01HG8QzLXU68cxJq5B3fqpYc


jvc56 (approved):

jvc56:
I approved before looking at the unit test failure. Not sure what's wrong, but it's probably some test that needs to be updated as opposed to a regression.

jvc56 (approved):

PR #526, opened 2026-04-26
Fully solve all etopk endgames

jvc56:


olaugh:
claude review notes

* PV display: TT-extended moves now labeled "exact" (endgame.c:417).** `pvline_extend_from_tt` now sets `negamax_depth = num_moves`, so TT-extended moves appear above the `---` separator. The extension follows `ttentry_move(tt_entry)` regardless of bound type — for `TT_LOWER`/`TT_UPPER` the stored move is typically a refutation, not a proven best. Suggest reverting this hunk so non-EXACT TT moves stay below the separator.

* Tested empirically: greedy-autoplay endgames at eplies=6, top_k=3, single-thread vs 8-thread on the same positions. Of 10 positions, 4 had the early-stop fire (depth 2/4/5/5 vs multi's 6); **all 4 produced the same best value as the multi-thread oracle**. Bug appears latent — didn't manifest on this sample. Worth either softening the comment or tightening the check (e.g. require `!use_heuristics`, or store an "all-leaves-terminal" bit alongside TT entries).

* Style: `multi_pv ? topk_n : best_value > -LARGE_VALUE` (endgame.c:2203) mixes int and bool. Clang-tidy 18's `readability-implicit-bool-conversion` may flag.

* TSAN clean on 3 multi-thread eplies=6 solves (no races, exit 0).

src/impl/endgame.c
```diff
@@ -375,6 +388,14 @@ void pvline_extend_from_tt(PVLine *pv_line, Game *game_copy,
       break;
     }
 
+    if (still_exact) {
+      int remaining_depth = max_depth - num_moves;
+      if (ttentry_flag(tt_entry) != TT_EXACT ||
+          ttentry_depth(tt_entry) < (uint8_t)remaining_depth) {
```
olaugh, line 394: counterproductive `(uint8_t)` cast on `remaining_depth` here, doesn't suppress any clang-tidy or cppcheck warning, actually generates *worse* code

Minimal reduction:

```c
bool test(uint8_t depth, int remaining_depth) {
  return depth < (uint8_t)remaining_depth;   // with cast
  // return depth < remaining_depth;         // without cast
}
```

arm64, clang and gcc-14, both -O2 and -O3 (cast persists through all optimization levels):

```
with cast:                without cast:
  cmp  w0, w1, uxtb         cmp  w1, w0
  cset w0, lo               cset w0, gt
```

The cast forces an extra `uxtb` (zero-extend byte) on `remaining_depth` and switches the comparison to unsigned. Without the cast, `depth` (uint8_t) just promotes to int and you get a single signed compare. Neither compiler recognizes that the input range makes the mask redundant.

Functionally equivalent for any `0 <= remaining_depth <= 255`, which always holds here (`ttentry_depth` is masked to 6 bits, max 63), but the cast is pure overhead — no warning, no perf, one extra instruction. Same nit applies to the matching cast on line 1766.

jvc56:
These changes have made the endgame solver nondeterministic. I'll need to get a better understanding of how endgame works before continuing.

olaugh:
Confirming the nondeterminism — I dug into this independently before seeing your comment and have some data that may help narrow down the mechanism.

**Depth distribution on test_topk_fully_solved case 1's position** (BGIV/DEHILOR, eplies=10, topk=3, 50 iters per thread count, seeds varied per iter):

| threads | d7 | d8 | d9 | d10 |
|---------|-----|-----|-----|------|
| 1 | 50 | 0 | 0 | 0 |
| 2 | 50 | 0 | 0 | 0 |
| 3 | 46 | 4 | 0 | 0 |
| 4 | 40 | 10 | 0 | 0 |
| 8 | 29 | 15 | 5 | 1 |

Single- and 2-threaded are deterministic; 3+ threads show varying `completed_depth` across runs. Score stayed at 11 in every run on this position so the nondeterminism is invisible via score, but `completed_depth` (and presumably PV / TT contents) is unstable.

**Likely mechanism:** the early-stop check at `iterative_deepening` lines 2561–2568 reads `any_leaf_game_unfinished`, which is a solver-wide atomic. Thread 0 resets it at the start of each ply and reads it at the end; in between, any worker that visits a depth-0 leaf with the game still ongoing sets it to 1. Whether that flag-set wins the race against thread 0's read varies with thread scheduling — that's the source of the nondeterminism.

The original implementation (commit 9684df2d) had a `(threads == 1)` guard on this check, with the comment:

> In multi-threaded mode ABDADA defers root moves to other threads; their leaves cannot be observed here, so the flag is unreliable and we fall back to ply == plies.

The test-adding commit (3f11a240) removed the guard. Re-adding it restores determinism (single-thread always fires when applicable; multi-thread always runs full eplies).

**Perf cost of restoring the guard** (case 1, threads=3):
- eplies=7: 0.60s → 0.60s (no change)
- eplies=8: 0.61s → 0.62s (+2%)
- eplies=9: 0.61s → 0.65s (+7%)
- eplies=10: 0.60s → 0.67s (+12%)

Cost scales with `eplies − actual_game_length`. For tightly-tuned eplies the guard is essentially free.

**What I tried but couldn't reproduce:** a score divergence (i.e., a "wrong answer" not just nondeterministic depth). I ran 360 multi-thread iterations across 9 different positions including some with score that varies per depth, and never observed a score that disagreed with the single-thread baseline. So the nondeterminism appears to be in `completed_depth` / search progress, not in the final returned value — at least on the positions I tried. Whether the race window can ever produce a wrong score is an open question; I just couldn't make it happen empirically.

Happy to share the stress test scaffolding if it's useful.

jvc56:
Fixed in #657

PR #547, opened 2026-06-07
tt: fastrange (multiply-shift) indexing for non-power-of-2 tables

olaugh:
Trying out the suggestion from Discord (RightBehindYou): index the transposition table with Lemire's multiply-shift ("fastrange") reduction instead of masking to a power of two, so the table can fill the requested memory fraction exactly.

Ref: https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/

## Change

- **Index:** `zval & size_mask` → `(uint64_t)(((__uint128_t)zval * num_entries) >> 64)` (one 64×64→128 multiply + shift). `num_entries` no longer needs to be a power of two.
- **Sizing:** `transposition_table_create` now uses the full `fraction_of_memory * total / 16` entry count (clamped to the platform minimum), instead of rounding **down** to the nearest power of two — which wasted up to ~50% of the budget. Bigger table within the same memory → better hit rate.

## The verification subtlety

The old scheme stored only 40 hash bits and reconstructed the full 64-bit hash from `40 stored bits + the index bits` — which worked because the index *was* the low `size_power` bits of the hash. fastrange derives the index from the **high** bits, so that no longer holds.

Fix: store the **low 40 bits** of the hash as a verification tag and compare directly on lookup. For any table with `≥ 2^(64−40) = 2^24` entries (the native minimum, `TT_MIN_SIZE_POWER`), each bucket spans `≤ 2^40` hashes, so two distinct hashes in the same bucket **cannot** share a tag — verification is exact, same as before. (WASM's `2^21` minimum loses a few high bits — the same tradeoff the old code already documented for small tables.)

The lockless (Hyatt 1999) key/data XOR torn-read detection and the separate power-of-2 `nproc` ABDADA table are unchanged.

## Validation

- `tt` test rewritten for the new scheme (asserts the bucket-index relationships directly); passes.
- **Endgame correctness** — `endgame` (multithreaded, 6 threads) and `eldar_v` exact-value solves pass, so TT pruning is unaffected.
- Clean under **ASan/UBSan** (new `malloc` sizing + the `__uint128` multiply).
- clang-format-20 + cppcheck clean. `__uint128_t` is already used in `bit_rack.h`, so it builds on all CI targets incl. WASM.

## Note

The tests clamp the TT to the `2^24` minimum (a power of two), so they don't exercise a non-power-of-2 size directly — the memory-utilization win shows up only with a real `-ttfraction`. Happy to add an endgame benchmark comparing hit rate / nodes-per-second before vs after if useful.

### DIFF
--- a/src/ent/transposition_table.h
+++ b/src/ent/transposition_table.h
@@ -19,6 +19,18 @@
 #define TT_MIN_SIZE_POWER 24 // 2^24 minimum (256 MB) for performance
 #endif
 
+// Number of low hash bits stored per entry as a verification tag. fastrange
+// indexing (see tt_bucket_index) derives the bucket from the HIGH hash bits, so
+// the LOW bits are what discriminate distinct positions within a bucket. As
+// long as the table has at least 2^(64 - TT_TAG_BITS) entries, every bucket
+// spans at most 2^TT_TAG_BITS hashes, so two distinct hashes in the same bucket
+// cannot share a tag -- verification is exact, matching the old power-of-2 +
+// index-reconstruction scheme. With TT_TAG_BITS == 40 that threshold is 2^24,
+// i.e. TT_MIN_SIZE_POWER on native builds (WASM's 2^21 minimum loses a few high
+// bits, the same tradeoff the old scheme accepted on small tables).
+#define TT_TAG_BITS 40
+#define TT_TAG_MASK ((UINT64_C(1) << TT_TAG_BITS) - 1)
+
 enum {
   TT_EXACT = 0x01,
   TT_LOWER = 0x02,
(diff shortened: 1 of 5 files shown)


olaugh:
Closing — benchmarking showed no worthwhile benefit for MAGPIE. At a 2x-table-size config (ttfraction 0.01: 16.8M vs 32M entries) fastrange only lowered the type-2 collision rate (45%→32%) with no change in hit rate (~27.5%), node count, or throughput (~1.47 M lookups/s), and the profile showed TT access inlines into a move-generation-dominated hot path so the index op (mask vs 128-bit multiply) is negligible. Not worth the added complexity / weaker 40-bit verification. The only real bug found along the way was the diagnostic TT counters overflowing atomic_int on billion-lookup searches; can resurface that as a tiny standalone fix if ever wanted.

PR #550, opened 2026-06-12
endgame: speed up first-win, with an adjustable depth-0 fallback cap

olaugh:
## Summary

The endgame's depth-0 fallback greedily evaluates **every** root candidate before the IDS loop, as a "best move if interrupted before depth 1 completes" safety net. At a 7-tile endgame root that is often **1000+ moves**, each costing a greedy leaf playout. For `first_win` (the path **pessimistic PEG** uses) this is almost pure overhead: first-win only needs win/loss, its depth-1 search is cheap, and the fallback value is overwritten the moment depth-1 completes (the common case).

This caps that sweep for `first_win`, via a new **adjustable** `EndgameArgs` field:

- `first_win_fallback_moves`: `0` = built-in default (12), `<0` = skip the sweep entirely, `>0` = evaluate at most this many top moves (moves are already sorted by estimate).

**Full-spread (non-first-win) solves — what the production endgame uses — are untouched.**

## Random-endgame benchmark (200 positions)

200 random nonstuck 7-tile endgames (self-play to bag-empty, seeded), first-win, 2-ply, 1 thread, quiet CPU:

| | total wall (200 pos) | avg/pos |
|---|---|---|
| uncapped fallback (old) | 11.70s | 58ms |
| **capped (default 12)** | **7.31s** | **37ms** |

**1.60× speedup.** Verdicts: 198/200 identical. The 2 that differ are **not regressions** — on both, the capped result is *closer to the exact full-spread value*. The uncapped fallback's greedy playouts seed the shared TT with shallow estimates that bias the narrow-window first-win search; capping removes that pollution:

| pos | exact (30-ply full-spread) | uncapped FW | capped FW |
|---|---|---|---|
| 44 | **+11** (win) | −9 (**wrong sign**) | **+9** (correct sign ✓) |
| 112 | **+14** | +16 (off by 2) | **+14** (exact ✓) |

So capping the fallback both speeds up first-win and removes a source of error in it.

## Single-leaf microbenchmark (representative 7-tile leaf, 1 thread, 2-ply, quiet CPU, 5 runs)

| first-win solve | wall |
|---|---|
| uncapped (old behavior — 1542 fallback playouts) | ~13ms |
| **capped (default 12)** | **~5ms** |
| skip fallback (`-1`) | ~5ms |

~2.6× on this leaf, identical result (+84). `capped == skip` confirms the 12-move fallback is negligible — the fallback is only consulted if IDS is interrupted before depth 1 (which doesn't happen here), so on a completed solve it's pure savings.

## Validation

- New `test_endgame_first_win_sign` (CI `test_table`): first-win must agree in **sign** with the exact full-spread solve, across the default and skip fallback caps.
- `endgame` + `endgamefirstwin` pass; release + dev (ASan/UBSan) builds clean; clang-format clean.

## Context (not changed here)

On the same leaf, first-win is at **parity with Macondo's first-win** (~5ms total each, same +84 value) — confirmed structurally: identical 1542 root moves, and MAGPIE's pruned KWG is actually *smaller* (4,907 vs 12,900 nodes, since MAGPIE minimizes the pruned GADDAG while Macondo serializes a trie). A residual ~1.5× on *full-spread* solves (MAGPIE ~26ms vs Macondo ~18ms) is the greedy-playout-vs-truncation leaf-eval difference — general endgame-core, not first-win, and out of scope for this PR.

### DIFF
--- a/src/impl/endgame.c
+++ b/src/impl/endgame.c
@@ -73,6 +73,11 @@ enum {
   CONSERVATION_VALUE_WEIGHT = 2,
   // Random noise range for thread jitter, centered around zero
   THREAD_JITTER_NOISE = 8,
+  // Max top-K ranked PVLines built for the per-ply display/callback.
+  MAX_RANKED_CALLBACK_PVS = 10,
+  // Default cap on the depth-0 interrupt-fallback sweep when first_win_optim is
+  // set: 0 = use this default, <0 = skip the sweep, >0 = explicit cap.
+  FIRST_WIN_D0_FALLBACK_MOVES = 12,
 };
 
 // Returns fraction of opponent's rack score that is stuck (0.0 = none, 1.0 =
@@ -2330,7 +2335,6 @@ static void build_ranked_pvs_and_notify(EndgameCtxWorker *worker, int depth,
                                         const PVLine *extended_pv,
                                         const SmallMove *initial_moves,
                                         int initial_move_count) {
-  enum { MAX_RANKED_CALLBACK_PVS = 10 };
   int n_ranked = initial_move_count < MAX_RANKED_CALLBACK_PVS
                      ? initial_move_count
                      : MAX_RANKED_CALLBACK_PVS;
@@ -2443,8 +2447,8 @@ void iterative_deepening(EndgameCtxWorker *worker, int plies) {
     d0_pv.num_moves = 0;
     d0_pv.negamax_depth = 0;
 
-    // first_win_fallback_moves: 0 = use this default, <0 = skip, >0 = explicit.
-    enum { FIRST_WIN_D0_FALLBACK_MOVES = 12 };
+    // first_win_fallback_moves: 0 = use FIRST_WIN_D0_FALLBACK_MOVES default,
+    // <0 = skip the sweep, >0 = explicit cap.
     int fw_fallback_cap = FIRST_WIN_D0_FALLBACK_MOVES;
     if (worker->solver->first_win_fallback_moves > 0) {
       fw_fallback_cap = worker->solver->first_win_fallback_moves;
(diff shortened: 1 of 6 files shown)


test/endgame_test.c
```diff
@@ -381,6 +381,66 @@ void test_endgame_dynamic_worker_injection(void) {
...
+        endgame_results_get_pvline(results, ENDGAME_RESULT_BEST)->score;
+    printf("endgame first_win=%d fallback=%d: value=%d\n", args.first_win,
+           args.first_win_fallback_moves, value);
+    if (args.first_win) {
+      assert(value < 0); // correct loss sign regardless of the fallback cap
+    } else {
+      assert(value == -63); // exact full-spread value
+    }
+    error_stack_destroy(error_stack);
+    endgame_ctx_destroy(ctx);
```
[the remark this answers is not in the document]
olaugh: Good catch — fixed. This test only calls `endgame_solve_inline()` (no `ErrorStack*` param), so the error_stack create/check/destroy was dead and the assert a no-op. Removed it. (It was copied from `run_injection_case`, where the stack is used by the spawned `endgame_solve` branch.)

src/impl/endgame.c
```diff
@@ -2464,7 +2443,18 @@ void iterative_deepening(EndgameCtxWorker *worker, int plies) {
     d0_pv.num_moves = 0;
     d0_pv.negamax_depth = 0;
 
-    for (int i = 0; i < initial_move_count; i++) {
+    // first_win_fallback_moves: 0 = use this default, <0 = skip, >0 = explicit.
+    enum { FIRST_WIN_D0_FALLBACK_MOVES = 12 };
```
jvc56: Unless we have a specific reason to declare this enum here, let's put it at the top of the file with the other enum values.

jvc56 (approved):

src/impl/endgame.c
olaugh: Nice catch, I've noticed Claude trying to do this before but usually made it fix it. Audited the codebase for other cases (there was only one more, in endgame.c). Fixed that and added a note to CLAUDE.md

PR #558, opened 2026-06-15
Add play chooser with delegation strategy and challenge decisions

olaugh:
## Summary

Adds a unified interface for getting a computer player's move that delegates to **static eval**, **Monte Carlo simulation**, or the **endgame solver** based on a configurable `PlayChooserStrategy`, plus a separate **challenge / no-challenge decision** that can be committed before the chooser knows its own follow-up play.

### New: `GameTimer` (`src/ent/game_timer.h`)
Per-side game clock (ported from the TUI branch's concept) so per-move budgets can be derived from a real game timer. Untimed games report `INFINITY` remaining.

### New: `PlayChooser` (`src/impl/play_chooser.{c,h}`)
- **Strategy struct** selects pre-endgame evaluation (static or sim) and endgame evaluation (static or endgame solver), so "static always", "static until endgame", and "sim + endgame solve" are all configurations.
- **Move budget** is a fixed value, derived from a `GameTimer` (remaining clock split across estimated remaining plays), or a default flat budget.
- **Challenge decision** values the *keep* vs *challenge-off* branches and never challenges a play the chooser believes is valid. On a successful challenge it records the offender's returned tiles in their known-rack-from-phonies (full rack when the bag is empty).
- **Endgame challenge path** solves both branches concurrently against **one shared transposition table** that also seeds the post-verdict move solve, and **exits early via a null-window resolve** once one branch is exact (game-over branches are exact for free; a finished branch interrupts its sibling, which then only proves which side of the verdict it falls on).

### Engine: caller-specified search window (`src/impl/endgame.{c,h}`)
Generalizes `first_win` into a fixed `[initial_alpha, initial_beta]` window (`use_initial_window`). A null window `[x, x+1]` proves whether the exact value exceeds `x`; `first_win` is now equivalent to a `[-1, 1]` window. Composes with the existing `first_win_fallback_moves` work on main.

## Tests (`./bin/magpie_test playchooser`)
- Game timer behavior.
- Keep-phony-for-3x3 (static and sim/timer variants).
- Challenge-off-and-block, with a naive-vs-informed contrast proving the rack knowledge is load-bearing.
- Three endgame challenge cases: keep for the only Q play (TRANQ through a phony I), typical challenge-off, and keep to deny a higher-scoring opponent replay.
- Fixed-window solve semantics (`[V, V+1]` fails low, `[V-1, V]` fails high against the exact value V).

## Validation
Release + full suite, ASAN/UBSAN, ThreadSanitizer (concurrent shared-TT solves), cppcheck, clang-tidy-18 (new files clean), clang-format-20, and circular-dependency check all pass. Rebased onto current `main`; the only merge points were the shared `EndgameArgs`/`EndgameCtx` field blocks alongside main's `first_win_fallback_moves`.


olaugh:
Superseded by #588, which combines this and #577 and makes the challenge decision evaluate each branch by its own game stage on one utility currency.

PR #570, opened 2026-06-23
Fix BitRack shift UB for alphabets larger than 32 letters

olaugh:
## Problem

A `BitRack` packs `BIT_RACK_BITS_PER_LETTER` (4) bits × `BIT_RACK_MAX_ALPHABET_SIZE` (32) letters into 128 bits, so it can only represent alphabets with **≤ 32 machine letters and ≤ 15 of any single letter** (blanks included) — exactly what `bit_rack_is_compatible_with_ld()` checks.

Three move-gen subsystems key on a BitRack: the **RIT cache**, the **KLV-leaves cache**, and the **WMP**. All three must be inert for a BitRack-incompatible alphabet — but nothing enforced that:

- WMP / RIT were inert only **by accident** — such lexicons ship no `.wmp` / rack-info-table files.
- The **KLV-leaves cache is always on** and built a BitRack via `bit_rack_create_from_rack()` with no compatibility check.

So for a larger alphabet (e.g. **Polish `OSPS49`, 33 letters**), the `ml = 32` loop iteration computes `shift = 128` and runs `bit_rack.high |= x << (shift - 64)` — a shift of 64 on a `uint64_t`, which is undefined behavior:

```
src/ent/bit_rack.h:116: runtime error: shift exponent 64 is too large for 64-bit type 'uint64_t'
```

Reached on the ordinary best-move path, **independent of WMP**:

```
#0 bit_rack_create_from_rack            bit_rack.h
#1 gen_look_up_leaves_and_record_exchanges move_gen.c   ← KLV-leaves cache
#2 generate_moves
#3 get_top_equity_move                  gameplay.c       (set -lex OSPS49 -wmp false)
```

On arm64 `x << 64` wraps the count mod 64 rather than dropping it, so beyond the UB the BitRack key can also *collide* — distinct racks comparing equal and returning a wrong cached leave value.

## Fix

Enforce the invariant centrally in `gen_load_position`:

- Cache `gen->bit_rack_compatible = bit_rack_is_compatible_with_ld(&gen->ld)`.
- When false, treat the **RIT** and **WMP** as absent (`NULL`) so neither builds a BitRack (`wmp_move_gen_init(NULL)` is the existing inert path).
- The always-on **KLV-leaves cache** consults the flag directly (it isn't gated by a nullable resource).
- Defensively bound the shift in `bit_rack_create_from_rack` (`else if (shift < 128)`) so a future ungated caller can never invoke UB rather than silently producing a corrupt key.

For compatible alphabets (≤ 32 letters, ≤ 15 each) this is a **no-op**: the gates resolve to the prior values and no `ml` reaches the new bound.

## Validation

- **Before/after** (dev ASAN/UBSAN, `UBSAN_OPTIONS=halt_on_error=1`): the `shift exponent 64` error fires without the fix; gone with it. Polish (`OSPS49`) and English (`CSW24`, which exercises the RIT/KLV cache) autoplay both run clean.
- New regression test `large_alphabet_movegen_test` generates `OSPS49` moves with a full rack + equity sort (the cache path). Full `movegen` and `bitrack` suites pass (`BUILD=release`).
- `format.py`, `cppcheck` (no errors), and circular-deps all clean.

Only alphabets with **more than 32 letters** (or >15 of a letter) are affected — Polish qualifies; English/French/Catalan fit — which is why it hadn't surfaced.

https://claude.ai/code/session_018LvFYMgVajp5e5U4m2rtaB

### DIFF
--- a/src/ent/bit_rack.h
+++ b/src/ent/bit_rack.h
@@ -112,9 +112,13 @@ static inline BitRack bit_rack_create_from_rack(const LetterDistribution *ld,
     const int shift = ml * BIT_RACK_BITS_PER_LETTER;
     if (shift < 64) {
       bit_rack.low |= (uint64_t)num_this << shift;
-    } else {
+    } else if (shift < 128) {
       bit_rack.high |= (uint64_t)num_this << (shift - 64);
     }
+    // A BitRack holds BIT_RACK_MAX_ALPHABET_SIZE letters; callers must gate on
+    // bit_rack_is_compatible_with_ld so ml never reaches the 128-bit boundary.
+    // The bound above keeps a misuse from shifting by >= 64 (undefined
+    // behavior) instead of silently producing a corrupt key.
 #endif
   }
   return bit_rack;
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -2721,6 +2721,7 @@ void gen_load_position(MoveGen *gen, const MoveGenArgs *args) {
   const Player *opponent = game_get_player(game, 1 - gen->player_index);
 
   memcpy(&gen->ld, game_get_ld(game), sizeof(LetterDistribution));
+  gen->bit_rack_compatible = bit_rack_is_compatible_with_ld(&gen->ld);
   gen->kwg = player_get_kwg(player);
   gen->kwg = (override_kwg == NULL) ? player_get_kwg(player) : override_kwg;
   const KLV *new_klv = player_get_klv(player);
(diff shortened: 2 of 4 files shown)


olaugh:
Superseded by #571 — the branch was renamed to `disable-bitrack-caches-for-incompatible-ld` to reflect the change (gating BitRack-keyed caches on letter-distribution compatibility), which auto-closed this PR. Same commit, updated description.

PR #571, opened 2026-06-23
Disable BitRack-keyed caches for BitRack-incompatible letter distributions

olaugh:
## What this changes

A `BitRack` is a 128-bit multiset key that packs `BIT_RACK_BITS_PER_LETTER` (4) bits × `BIT_RACK_MAX_ALPHABET_SIZE` (32) letters. It can therefore only represent a letter distribution with **≤ 32 machine letters and ≤ 15 of any single letter** (blanks included) — precisely the `bit_rack_is_compatible_with_ld()` predicate.

Three move-generation subsystems use a BitRack as a key — the **RIT cache**, the **KLV-leaves cache**, and the **WMP**. The invariant should be: *for a BitRack-incompatible distribution, none of them engage.* Nothing enforced it. `bit_rack_is_compatible_with_ld()` was referenced only in a unit test; WMP and RIT stayed inert for large alphabets purely because such lexicons ship no `.wmp` / rack-info-table files.

This PR makes the invariant explicit and enforced centrally in `gen_load_position`:

- Cache `gen->bit_rack_compatible = bit_rack_is_compatible_with_ld(&gen->ld)` once per position.
- When false, treat the **RIT** and **WMP** as absent (`NULL`), so neither builds a BitRack (`wmp_move_gen_init(NULL)` is the existing inactive path).
- The **KLV-leaves cache** is always available (not gated by a nullable resource), so it consults the flag directly.

For a compatible distribution (every common lexicon — English, French, Catalan, …) this is a **no-op**: the gates resolve to the prior values.

## The bug it fixes

The KLV-leaves cache (`gen->klv_leaves_cache`) is on for any equity-sorted generation with a full rack and built its BitRack key **unconditionally**. For an incompatible alphabet (e.g. **Polish `OSPS49`, 33 letters**), the `ml = 32` iteration of `bit_rack_create_from_rack` runs `bit_rack.high |= x << (shift - 64)` with `shift = 128` — a shift of 64 on a `uint64_t`, undefined behavior:

```
src/ent/bit_rack.h: runtime error: shift exponent 64 is too large for 64-bit type 'uint64_t'
```

It is reached on the ordinary best-move path, **independent of the WordMap**:

```
#0 bit_rack_create_from_rack             bit_rack.h
#1 gen_look_up_leaves_and_record_exchanges  move_gen.c   ← KLV-leaves cache
#2 generate_moves
#3 get_top_equity_move                   gameplay.c       (set -lex OSPS49 -wmp false)
```

On arm64 `x << 64` wraps the shift count mod 64 rather than dropping the bits, so beyond the UB the key can also *collide* — two distinct racks comparing equal and returning a wrong cached leave value.

## Validation

- **Before/after** under dev ASAN/UBSAN with `UBSAN_OPTIONS=halt_on_error=1`: the shift error fires without the fix and is gone with the central gate alone. Polish (`OSPS49`) and English (`CSW24`, which exercises the RIT/KLV cache) autoplay both run clean.
- New regression test `large_alphabet_movegen_test` generates `OSPS49` moves with a full rack + equity sort (the cache path). The full `movegen` and `bitrack` suites pass (`BUILD=release`).
- `format.py`, `cppcheck` (no errors), and circular-deps all clean.

Only distributions with **more than 32 letters** (or more than 15 of a letter) are affected — Polish qualifies; the common lexicons fit — which is why it had not surfaced.

(Supersedes #570, which auto-closed when its branch was renamed.)

https://claude.ai/code/session_018LvFYMgVajp5e5U4m2rtaB

### DIFF
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -2721,6 +2721,7 @@ void gen_load_position(MoveGen *gen, const MoveGenArgs *args) {
   const Player *opponent = game_get_player(game, 1 - gen->player_index);
 
   memcpy(&gen->ld, game_get_ld(game), sizeof(LetterDistribution));
+  gen->bit_rack_compatible = bit_rack_is_compatible_with_ld(&gen->ld);
   gen->kwg = player_get_kwg(player);
   gen->kwg = (override_kwg == NULL) ? player_get_kwg(player) : override_kwg;
   const KLV *new_klv = player_get_klv(player);
@@ -2751,7 +2752,10 @@ void gen_load_position(MoveGen *gen, const MoveGenArgs *args) {
   gen->klv = new_klv;
   gen->klv_mutation_counter_at_load = new_klv_mutation_counter;
   gen->klv_instance_fp_at_load = new_klv_instance_fp;
-  const RackInfoTable *new_rit = player_get_rack_info_table(player);
+  // The RIT keys on a BitRack, so it is unusable for a BitRack-incompatible
+  // alphabet; treat it as absent (the WMP is gated the same way below).
+  const RackInfoTable *new_rit =
+      gen->bit_rack_compatible ? player_get_rack_info_table(player) : NULL;
   if (new_rit != gen->rack_info_table) {
     memset(gen->rit_cache_valid, 0, sizeof(gen->rit_cache_valid));
     // Leave values are stored in the subrack cache; they depend on the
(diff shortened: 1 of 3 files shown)


olaugh:
Thanks @copilot — addressed:

- **`move_gen.h` field comment** — rewritten to state both constraints (≤ 32 letters *and* ≤ 15 of any one letter, blanks included, per `bit_rack_is_compatible_with_ld`) and to note that the **WMP** is gated by this flag too, alongside the RIT and KLV-leaves caches.
- **`move_gen_test.c` assertion** — good catch on the bounded top-equity heap. Changed to assert that generation produced at least one move (scoring + non-scoring) rather than that a non-scoring play survived eviction.
- **`bit_rack.h` defensive shift bound** — this comment refers to a defensive `shift < 128` bound that has since been **removed** (the central gate makes the primitive unreachable for incompatible distributions, so the extra per-iteration branch in the hot path wasn't worth it). `bit_rack.h` is no longer touched by this PR.

jvc56 (approved):

PR #577, opened 2026-07-05
PEG pre-endgame evaluation for PlayChooser challenge decisions

olaugh:
## What

Adds `PLAY_CHOOSER_EVAL_PEG` so the pre-endgame (PEG) solver can value the two branches of a challenge decision — **keep** (phony stands, opponent draws) vs **challenge** (phony off, opponent loses the turn) — letting the pre-endgame make challenge-or-not calls, not just move choices. Stacked on `play-chooser`.

## How it works

**Score+win utility.** A branch is valued by the simmer's `sim_utility_blend` verbatim: `(w_winpct·win% + w_spread·σ(spread/scale)) / (w_winpct + w_spread)`, bounded `[0,1]`. So margin separates would-be win% ties (a certain-win keep branch is discounted below 1.0 by its finite margin), and endgame/game-over branches map their decided spread through the same sigmoid. Exposed via `utility_w_winpct` / `utility_w_spread` / `utility_spread_scale` on the strategy; zero/unset defaults match the simmer (pure win%).

**Greedy seed for the branch valuation.** Challenge/keep branches use PEG's greedy seed (new `PegArgs.greedy_seed_only`): full scenario enumeration with the deterministic playout and **no open-ended deep endgame solves**, so the value is bounded and machine-independent once the seed completes, rather than depending on how many refinement stages a wall-clock budget happened to finish. Actual move selection (`choose_move`) keeps the full cascade for accuracy.

**Parallel arms.** Both branches are valued concurrently, each with the whole decision budget and its own thread slice + endgame scratch — mirroring the endgame challenge decider.

**Default to challenge.** A detected phony is challenged by default; the chooser keeps it only with **two valid branch evaluations** where keeping is strictly better. Branch results carry an explicit `valid` flag rather than overloading `0.0`, so a severe budget that yields no usable result (or a failed solve) falls back to challenging rather than gifting an unchallenged invalid play. The endgame challenge decider follows the same rule (validity-aware compare + null-window resolve defaults to challenge if the proof doesn't complete).

## Tests

`test_peg_challenge_decision` (real 4-in-bag CSW24 position): the same phony hook yields opposite verdicts — **OBEATY** is challenged, **CBEATY** is kept (challenging would hand the opponent a do-over with its strong rack). Asserts the score+win tiebreak (CBEATY keep < 1.0) and the degenerate default (1µs budget → both branches invalid → challenge). `playchooser` + full suite pass under `-Werror` + ASan/UBSan.

## Self-review notes

- **Thread safety:** the two PEG arms share no mutable state — each `run_peg` solve is self-contained (own `ThreadControl`, own `PegResult`), and only the keep arm can reach the empty-bag endgame path (the challenge arm always retains bag tiles), so `get_endgame_tt`'s lazy init and the per-arm endgame ctx are never touched concurrently.
- **Perf notes captured during development:** greedy-vs-cascade is bit-identical for these positions; parallel vs sequential is a wash for the compute-bound greedy path (core-seconds conserved) and pays off mainly when the keep arm collapses to a deep endgame; severe budgets (<~1s/arm) degrade to the challenge default.
- Docs on the enum / `ChallengeDecision` units / strategy fields updated to describe the utility.

https://claude.ai/code/session_01FKQniDvyWewhbMhuCrFKS2


olaugh:
Superseded by #588, which combines this (PEG pre-endgame eval) and #558 and unifies the challenge decision across all six keep/challenge game-stage combinations.

PR #584, opened 2026-07-09
Autoanalyze upgrades

jvc56:


src/impl/config.c
```diff
@@ -7913,17 +7936,94 @@ static void config_fill_analyze_args(Config *config, AnalyzeArgs *analyze_args,
...
+      pos = end;
+      ok = error_stack_is_empty(parse_error_stack) &&
+           consume_literal(&pos, " aeql=");
+    }
+    if (ok) {
+      *aeql = string_to_double_prefix(pos, &end, parse_error_stack);
+      pos = end;
+      ok = error_stack_is_empty(parse_error_stack) &&
+           consume_literal(&pos, " ===");
+    }
```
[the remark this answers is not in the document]
jvc56: Done.

src/impl/config.c
```diff
@@ -3109,8 +3115,11 @@ void config_fill_endgame_args(Config *config, EndgameArgs *endgame_args) {
   endgame_args->enable_pv_display = true;
   endgame_args->per_ply_callback = NULL;
   endgame_args->per_ply_callback_data = NULL;
-  endgame_args->soft_time_limit = 0;
-  endgame_args->hard_time_limit = 0;
+  const double endgame_time_limit = config->endgame_time_limit_seconds != 0
```
olaugh: **Silent behavior change worth calling out in the PR description**: `soft/hard_time_limit` previously hardcoded 0 (unlimited); with this fallback, every existing `endgame` command invocation now inherits `-tlim`, whose default is **60 seconds**. Deep endgame solves that used to run to completion will silently stop at 60s and can return shallower results. If the intent was only to give autoanalyze a bounded endgame, consider keeping the fallback at 0 for the plain endgame path (or defaulting `-etlim` to 0-means-unlimited rather than 0-means-tlim). Tests are unaffected since they bypass config fill.

src/impl/config.c
```diff
@@ -2196,6 +2211,7 @@ char *impl_help(Config *config, ErrorStack *error_stack) {
         ARG_TOKEN_CUTOFF,                  /* cutoff */
         ARG_TOKEN_ENDGAME_PLIES,           /* eplies */
         ARG_TOKEN_ENDGAME_TOP_K,           /* etopk */
+        ARG_TOKEN_ENDGAME_TIME_LIMIT,      /* etlim */
```
olaugh, line 2213: Nit: alphabetical ordering — "etlim" sorts before "etopk" ('l' < 'o'), so this entry should come one line earlier. (`pegtlim` is placed correctly between `pegstride` and `pegtopk`.)

src/impl/config.c
```diff
@@ -7913,17 +7936,104 @@ static void config_fill_analyze_args(Config *config, AnalyzeArgs *analyze_args,
...
+  return true;
+}
+
+// Reads report_path and, if it ends in a "=== Analysis Complete: ..."
+// trailer (written unconditionally by analyze_game on a clean run), parses
+// the turn count and clamped WPL/EqL/AEqL totals out of it. Returns false
+// (leaving the outputs untouched) if the file is missing or has no such
+// trailer, e.g. because it doesn't exist yet or a previous run crashed or
+// errored partway through.
+static bool read_report_completion_stats(const char *report_path, int *turns,
```
olaugh, line 7986: Following up on Copilot's comment here: agreed on the contract mismatch, though it's not a functional bug today — both call sites initialize the outputs and only consume them when this returns true. Cheapest real fix is to parse into locals and assign the out-params only after the final `consume_literal(&pos, " ===")` succeeds; that makes the documented "leaving the outputs untouched" true instead of narrowing the doc comment.

src/impl/peg.c
```diff
@@ -1988,6 +1985,51 @@ static int peg_select_survivors(PegRankedCand *ranked, int live_count, int keep,
...
 
+// Moves any protected candidate to the front of ranked[0..count). A stage's
+// live-mode evaluation (see the sequential per-candidate loop in peg_solve)
+// scores candidates strictly in array order and stops as soon as the
+// deadline passes, so whatever peg_select_survivors appended after the
+// top-K cut is evaluated last and is the first thing dropped under a tight
+// time budget. Moving it to the front instead guarantees it gets scored
+// before the deadline can cut the stage short, mirroring how the endgame
+// solver pins its protected actual move to root index 0.
+static void peg_force_protected_to_front(PegRankedCand *ranked, int count,
```
olaugh, line 1996: **Latent bug with `n_protect > 1`**: this loop doesn't `break` after swapping, so a *second* protected match swaps the first protected move back out of position 0 — only the last protected move found is guaranteed front placement, and an earlier one can get moved from inside the top-K region out to a straggler slot. Harmless today (analyze passes exactly one protected move), but the API accepts N. Either `break` after the first swap, or compact all protected moves to the front (`ranked[write_idx++]`-style) if the intent is to front-load all of them. Same issue in `peg_force_protected_move_ptrs_to_front` below.

src/impl/random_variable.c
```diff
@@ -555,9 +555,11 @@ double rv_sim_sample(RandomVariables *rvs, const uint64_t play_index,
   }
   sim_results_increment_iteration_count(sim_results);
 
-  return sim_utility_blend(wpct, spread, simmer->utility_w_winpct,
-                           simmer->utility_w_spread,
-                           simmer->utility_spread_scale);
+  const double utility =
+      sim_utility_blend(wpct, spread, simmer->utility_w_winpct,
+                        simmer->utility_w_spread, simmer->utility_spread_scale);
+  simmed_play_add_utility_stat(simmed_play, utility);
```
olaugh: Perf note on the sim hot path: this adds a mutex lock/unlock + stat push per sample even when `utility_w_spread == 0`, where BU is never displayed and the utility equals the win-pct blend anyway. Given how hot `rv_sim_sample` is, consider gating the push on `simmer->utility_w_spread > 0` (or folding it into an existing critical section) so the default-off case keeps the inner loop unchanged.

src/impl/analyze.c
```diff
@@ -1301,4 +1475,33 @@ void analyze_game(AnalyzeArgs *analyze_args, AnalyzeCtx **analyze_ctx,
...
+                          &equity_lost, &adjusted_equity_lost);
+      total_win_pct_lost += win_pct_lost;
+      total_equity_lost += equity_lost;
+      total_adjusted_equity_lost += adjusted_equity_lost;
+    }
+    char *trailer = get_formatted_string(
+        "\n=== Analysis Complete: turns=%d wpl=%.2f eql=%.2f aeql=%.2f ===\n",
+        ctx->turn_results_count, total_win_pct_lost, total_equity_lost,
+        total_adjusted_equity_lost);
+    append_string_to_file(report_path, trailer, error_stack);
```
olaugh, line 1511: Two documentation-worthy consequences of the trailer (both look deliberate, just worth a line in the docs/PR description): (1) it's appended to CSV-mode reports too, so downstream CSV consumers must tolerate a non-CSV footer; (2) since directory-mode skip keys only off the trailer's existence, reports produced under *different* analysis settings (plies, plays, budgets) still count into tournament aggregates, and the only way to force a refresh is deleting the report files.

olaugh (commented):
Did a full pass over the diff (all 16 files), with the load-bearing details verified against the head branch. Overall this is a strong PR — the exact actual-move evaluation inside the same endgame solve is a real correctness upgrade over the old second sign-flipped solve, and the alpha-window reasoning behind pinning it to root index 0 checks out (first root move always gets a full window in PVS; the displaced best move still fails high and gets re-searched exactly).

Things I verified that are sound and deserve a mention:
- The `compare_moves_without_equity(...) == -1` idiom looks alarming but is correct — that comparator's documented convention is -1 = equivalent.
- The exchange `row/col/dir` zeroing in validated_move.c matches movegen's literal `0, 0, 0` for exchanges, so both producers now compare equal. This soundly underpins the similarity-key → direct-compare switch.
- `AnalyzeCtx` is calloc'd so the embedded `PegResult` starts zeroed, and `peg_result_destroy` is zero-safe — destroy-before-solve is fine on the first turn.
- All other `EndgameArgs` call sites memset or `{0}`-init, so the new `actual_move` field is NULL everywhere except analyze, which sets and clears it explicitly.

Two review-level items that aren't line-specific:
1. **This branch is stacked on `sigmoid-uspread-default` (#582)** — merging this first will silently carry in #582's uspread=0.5 default (and make the new BU column appear in default sim output). Suggest merging #582 first or noting the dependency in the description.
2. **Test coverage**: the actual-move exact-value path (subtle alpha-beta invariant), the single-tile augmentation branch, and the trailer parse/skip logic all land without tests. A cheap one following the endgame_test.c patterns: solve a known position with `actual_move` set, assert `ENDGAME_RESULT_ACTUAL` matches a second full solve of that move.

Line comments follow for the rest. Re Copilot's comment on `read_report_completion_stats`: it's right about the contract/implementation mismatch, though both current call sites init locals and only consume on `true`, so it's not a functional bug today — see my inline suggestion.

olaugh (approved):

PR #588, opened 2026-07-13
Play chooser: per-branch, stage-aware challenge decisions on one utility (combines #558 + #577)

olaugh:
## Summary

Combines #558 (play chooser) and #577 (PEG pre-endgame eval) and unifies the challenge decision so **each branch is evaluated by the method for its own game stage**, all projected onto one comparable currency. Supersedes #558 and #577.

## The problem

`play_chooser_decide_challenge` weighs two hypothetical positions — the **keep** branch (phony stands; the opponent draws replacement tiles) and the **challenge** branch (phony comes off; opponent loses the turn, no draw). Only the keep branch draws, so the branches can be in **different stages**: keep's bag = `pre − tiles_played` (can hit an endgame), challenge's bag = `pre` (unchanged). The old code chose **one** eval mode from the pre-move phase and forced it on both, and the three modes returned three different currencies (SIM raw win%, PEG `[0,1]` utility, STATIC/ENDGAME spread) — so a keep branch that emptied the bag was approximated with SIM/PEG instead of solved exactly.

## The change

Each branch is now valued with the method for **its own** bag size and projected onto the shared score+win utility (`sim_utility_blend`, driven by `uwin`/`uspread`/`uspreadscale` from #582/#585), so the branches stay comparable across all six stage pairs:

| | keep→ | SIM | PEG | EG |
|---|---|---|---|---|
| **challenge** | SIM | ✓ | ✓ | ✓ |
| | PEG | — | ✓ | ✓ |
| | EG | — | — | ✓ |

(The keep branch is always the same or a later stage, so those six are the reachable pairs.)

- **`sim_results`**: the SIM branch needs the best play's mean utility, but previously discarded spread and returned raw win%. #584 (now merged) added the per-play `utility_stat` — the blended sample BAI averages — plus its getter and a utility-aware `compare_simmed_plays`; this PR builds directly on that, adding only `sim_results_get_best_move_utility` for an **exact** readback of the best play's mean utility (no Jensen gap, uses the absolute spread the simmer sigmoids). The SIM branch reports that value.
- **`play_chooser`**: `run_sim` now runs under the strategy's utility weights (was hardcoded pure win%); `evaluate_position` is a single per-stage→utility evaluator (STATIC keeps spread as the no-win-model fallback); `decide_challenge` dispatches per branch. The **EG/EG** case keeps the concurrent shared-TT null-window endgame decider — for two decided positions its higher-spread verdict equals the higher-utility one (utility is monotonic in spread), so the optimization is preserved with no behavior change.

## Tests

`test_challenge_stage_combinations` exercises all six stage pairs end to end on real fixtures (SEALED_CLUSTER 1-tile phony for the full-bag cases; the real bag-4/5 PEG fixture with 1/4/5-tile phonies for the low-bag and endgame-keep cases), asserting each branch reports the shared `[0,1]` utility (spread for the EG/EG decider).

## Validation

- Up to date with `main` (including #584); the overlapping `utility_stat` was reconciled onto #584's now-canonical version, keeping only `sim_results_get_best_move_utility` on top.
- `playchooser` (all six stage pairs + a single-threaded decider case), `sim`, `peg`, `endgame`, `movegen` green at `-O0`/ASAN and `-O3`.
- `format.py`, `cppcheck`, `clang-tidy` clean.

https://claude.ai/code/session_01FKQniDvyWewhbMhuCrFKS2


src/impl/play_chooser.c
```diff
@@ -0,0 +1,1007 @@
...
+
+  // Each branch is evaluated with the method for ITS OWN stage: the keep branch
+  // draws to full and can fall to an endgame (or low-bag pre-endgame) while the
+  // challenge branch (a lost turn, no draw) keeps its bag, so the two can be in
+  // different stages. Every stage projects onto the same [0, 1] utility, so the
+  // branches remain comparable.
+  const play_chooser_eval_t keep_eval =
+      play_chooser_get_eval_for_phase(strategy, keep_game);
+  const play_chooser_eval_t challenge_eval =
+      play_chooser_get_eval_for_phase(strategy, challenge_game);
```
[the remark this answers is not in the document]
olaugh: Fixed in 4803c3d7. STATIC reports spread, not the [0,1] utility, so it isn't comparable to a sibling utility. If either branch resolves to STATIC, both branches are now valued with STATIC (comparable spreads). This is only reachable without win percentages; with them a pre-endgame branch is SIM/PEG and an empty-bag branch is ENDGAME, all on the utility scale.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1007 @@
...
+                    &keep_branch);
+    play_chooser_solve_endgame_branch(&challenge_branch);
+    cpthread_join(keep_thread);
+    thread_control_destroy(keep_thread_control);
+    thread_control_destroy(challenge_thread_control);
+    if (!error_stack_is_empty(keep_error_stack)) {
+      error_stack_push(
+          error_stack, error_stack_top(keep_error_stack),
+          string_duplicate("keep branch challenge evaluation failed"));
+    }
```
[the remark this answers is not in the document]
olaugh: Fixed in 4803c3d7. The keep branch's message is now preserved: I take its top error code, pull the detailed string via error_stack_get_string_and_reset, and push a formatted message that includes it (freeing the intermediate string) instead of dropping it for a generic one.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1007 @@
...
+PlayChooser *play_chooser_create(const PlayChooserStrategy *strategy) {
+  if (strategy->pre_endgame_eval == PLAY_CHOOSER_EVAL_ENDGAME) {
+    log_fatal("play chooser pre-endgame evaluation cannot be the endgame "
+              "solver");
+  }
+  if ((strategy->pre_endgame_eval == PLAY_CHOOSER_EVAL_SIM ||
+       strategy->endgame_eval == PLAY_CHOOSER_EVAL_SIM) &&
+      strategy->win_pcts == NULL) {
+    log_fatal("play chooser sim evaluation requires win percentages");
+  }
```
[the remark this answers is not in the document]
olaugh: Fixed in 4803c3d7. play_chooser_create now log_fatals when endgame_eval is neither the endgame solver nor STATIC, so endgame_eval=PEG/SIM no longer silently degrades or runs a pointless empty-bag rollout.

src/impl/endgame.c
```diff
@@ -627,6 +630,9 @@ void endgame_ctx_reset(EndgameCtx *es, EndgameResults *results,
                        const EndgameArgs *endgame_args) {
   es->first_win_optim = endgame_args->first_win;
   es->first_win_fallback_moves = endgame_args->first_win_fallback_moves;
+  es->initial_window_optim = endgame_args->use_initial_window;
+  es->initial_window_alpha = endgame_args->initial_alpha;
+  es->initial_window_beta = endgame_args->initial_beta;
```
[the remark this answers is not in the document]
olaugh: Fixed in 4803c3d7. endgame_ctx_reset now sets initial_window_optim = use_initial_window && !first_win, making first_win take precedence explicitly so every downstream flag check (aspiration disabling, etc.) sees a consistent state instead of an ambiguous both-set combination.

src/impl/random_variable.c
```diff
@@ -555,14 +555,17 @@ double rv_sim_sample(RandomVariables *rvs, const uint64_t play_index,
...
+  // Record the exact sample BAI averages so the arm's mean utility is readable.
+  simmed_play_add_utility_stat(simmed_play, utility);
+  return utility;
 }
 
 static int rv_sim_get_best_arm_index(const RandomVariables *rvs) {
   const Simmer *simmer = (const Simmer *)rvs->data;
-  return sim_results_get_best_move_index(simmer->sim_results);
+  return sim_results_get_best_win_pct_index(simmer->sim_results);
 }
```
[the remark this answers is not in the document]
olaugh: Thanks — traced this. Three points:

1. **Behavior-preserving rename, not a change.** `rv_sim_get_best_arm_index` called `sim_results_get_best_move_index` before; I only renamed that function to `sim_results_get_best_win_pct_index`. The win%-best selection here is unchanged, and the win%-vs-utility inequality already existed for every utility-blended sim since #582/#585 (default `uspread=0.5`), independent of this PR.

2. **Impact is a sampling-count heuristic, not the winner.** The callback result is stored as `avoid_prune_best_arm_idx` and used only in `sim_unpruned_to_winner` as the top-up target (`arm_data[avoid_prune_best_arm_idx].num_samples`, bai.h:466-467) — i.e. how many extra samples to draw to level the other arms before finalizing. The *final* move is still utility-ranked: `sim_results_get_best_move` → `sim_results_get_best_utility_index` (BAI best arm when `uspread!=0`). So a mismatch here changes the top-up sample budget, not which play is selected.

3. **A correct fix is non-trivial and out of scope here.** Mid-sim there is no ready utility-best: `bai_result`'s best arm isn't set until the sim ends (which is exactly why `sim_results_get_best_utility_index` has to fall back). Ranking by the per-arm mean utility mid-sim (e.g. via the new `utility_stat`) would change BAI avoid-prune sampling for *all* utility sims and needs its own sim-strength validation, so it belongs in a focused simmer change rather than this play-chooser unification.

Leaving the callback as-is here (resolving on that basis). Happy to open a separate issue/PR to make the avoid-prune winner utility-consistent if you think the sampling change is worth validating.

olaugh:
Comparing the `SimmedPlay.utility_stat` overlap with #584 (flagged on Discord), hunk by hunk:

## Identical in both (textual conflicts, trivial to resolve)

Both PRs add the same core, nearly byte-for-byte: `Stat *utility_stat` on `SimmedPlay` (same position), identical create/reset/destroy wiring, an identical mutex-guarded `simmed_play_add_utility_stat()`, an identical `simmed_play_get_utility_stat()` accessor, and the **same `rv_sim_sample` rewrite** (capture the `sim_utility_blend` value, push it, return it). Whoever merges second just drops their duplicate hunks.

## Real differences

1. **Display sort.** #584 changes `compare_simmed_plays`: when `utility_w_spread > 0`, the *displayed* leaderboard ranks by mean BU (pass tiebreak factored out), so the display order matches autoplay's selection. #588 leaves the display sort as win%+equity and instead documents the distinction by renaming `sim_results_get_best_move_index` → `sim_results_get_best_win_pct_index`. These collide directly: after #584, #588's comment "the display sort does NOT consult the utility" becomes false, and the rename conflicts with #584 keeping the old name.
2. **Per-play mode flag.** #584 stamps `utility_w_spread` onto every `SimmedPlay` (refreshed alongside `cutoff`) with a `..._is_set()` predicate — needed because its comparator only sees `SimmedPlay*`. #588 keeps the weight solely on `SimResults`. If #584's BU-sorted display survives the merge, its per-play stamp comes with it.
3. **`simmed_play_copy`.** #584 copies `utility_stat` in `simmed_play_copy`; **#588 does not** — that's a gap in this PR regardless of merge order: the deep-copy path (e.g. saved/resumed SimResults) would silently drop the utility samples. Should be added here even if #584 merges first.
4. **Unique consumers.** #584 layers a BU display column (sim_string.c) and autoanalyze usage on top. #588's unique piece is `sim_results_get_best_move_utility()` — exact readback of the best arm's mean BAI utility for PlayChooser branch values (avoiding the Jensen gap of re-blending summary stats). These don't conflict with each other.
5. **Selection refactor.** #588 restructures `sim_results_get_best_move` into a static `get_best_utility_index` — functionally identical to current main (post-#585), so it's cosmetic; #584 doesn't touch selection.

## Suggested reconciliation

Land #584 first (older, and its comparator change is a UX fix worth having); then this PR rebases, drops the duplicated core, keeps `sim_results_get_best_move_utility()`, adapts or drops the `best_win_pct_index` rename to match #584's comparator world, and adopts the `simmed_play_copy` line. One more heads-up for whoever rebases last: draft #589 also rewrites the same `rv_sim_sample` return site (leftover-aware blend spread), so all three PRs meet on those lines.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

src/ent/sim_results.c
```diff
@@ -743,4 +743,26 @@ const Move *sim_results_get_best_move(const SimResults *sim_results) {
...
+        bai_result_get_best_arm(sim_results_get_bai_result(sim_results));
+    if (best_play_idx < 0) {
+      best_play_idx = sim_results_get_best_move_index(sim_results);
+    }
+  }
+  if (best_play_idx < 0) {
+    return 0.0;
+  }
+  return stat_get_mean(simmed_play_get_utility_stat(
+      sim_results_get_simmed_play(sim_results, best_play_idx)));
```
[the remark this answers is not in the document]
olaugh: Good catch — real regression from the #584 reconciliation. Fixed in db94: when `utility_w_spread == 0.0` the function now reads `simmed_play_get_win_pct_stat` (the utility *is* the win% on that path, and `utility_stat` is deliberately left empty by the simmer). Added a pure-win% SIM/SIM combo case that asserts both branch utilities are strictly in (0, 1), which a plain [0, 1] range check would not catch.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1068 @@
...
+  // keeps the phony only when the keep branch is strictly better, so a tie
+  // challenges: when resolving the challenge branch the window sits just
+  // below the exact keep value (challenge wins if it merely matches it),
+  // and when resolving the keep branch it sits just above the exact
+  // challenge value (keep must strictly exceed it to avoid the challenge).
+  double remaining_seconds =
+      decision_seconds - ctimer_elapsed_seconds(&decision_timer);
+  if (remaining_seconds < PLAY_CHOOSER_MIN_RESOLVE_SECONDS) {
+    remaining_seconds = PLAY_CHOOSER_MIN_RESOLVE_SECONDS;
+  }
```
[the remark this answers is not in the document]
olaugh: Fixed — when `remaining_seconds <= 0.0` (budget already spent) the resolve is now skipped and the decision defaults to challenge, rather than clamping up to `PLAY_CHOOSER_MIN_RESOLVE_SECONDS` and running an extra window past the deadline.

src/ent/sim_results.c
```diff
@@ -743,4 +743,32 @@ const Move *sim_results_get_best_move(const SimResults *sim_results) {
   }
   return simmed_play_get_move(
       sim_results_get_simmed_play(sim_results, best_play_idx));
+}
+
+// Mean utility (win%+spread blend in [0, 1]) of the sim's best play, using the
+// same best-play selection as sim_results_get_best_move. Not thread safe,
+// assumes the sim is finished. Returns 0 if there are no plays.
+double sim_results_get_best_move_utility(const SimResults *sim_results) {
```
jvc56, line 737: Both sim_results_get_best_move_utility and sim_results_get_best_move should call sim_results_get_best_move_index, since all of them should be using utility at this point. I think the following comment:

`Falling through to sim_results_get_best_move_index would re-rank by raw
  // win_pct_stat, ignoring the blend entirely`

is outdated as the move compare function now uses blended utility when available.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1076 @@
...
+  const int sim_plies = strategy->sim_plies > 0
+                            ? strategy->sim_plies
+                            : PLAY_CHOOSER_DEFAULT_SIM_PLIES;
+  ThreadControl *thread_control = thread_control_create();
+  thread_control_set_status(thread_control, THREAD_CONTROL_STATUS_STARTED);
+
+  const int player_on_turn_index = game_get_player_on_turn_index(game);
+  const Player *opponent = game_get_player(game, 1 - player_on_turn_index);
+
+  SimArgs sim_args = {0};
```
jvc56: We should prefer to use sim_args_fill here so that when new fields are added to the SimArgs the compiler will catch anything we've missed.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1076 @@
...
+  sim_args.bai_options.num_threads = num_threads;
+  sim_args.bai_options.cutoff = 0.0;
+  sim_args.bai_options.parent_worker_thread_index = 0;
+  sim_args.bai_options.arm_avoid_prune = NULL;
+  sim_args.bai_options.num_arm_avoid_prune = 0;
+  // Run the sim under the chooser's utility weights so it ranks by (and
+  // records) the same win%+spread blend used as the branch's value below.
+  sim_args.utility_w_winpct = play_chooser_util_w_winpct(strategy);
+  sim_args.utility_w_spread = strategy->utility_w_spread;
+  sim_args.utility_spread_scale = play_chooser_util_spread_scale(strategy);
```
jvc56: These (and other fields as required) should be added as args to the sim_args_fill function so that this code can call that function instead of setting SimArgs directly.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1076 @@
...
+    int num_threads, double budget_seconds,
+    ThreadControl *external_thread_control, bool use_window,
+    int32_t window_alpha, int32_t window_beta, Move *out_move,
+    int32_t *out_value, ErrorStack *error_stack) {
+  ThreadControl *thread_control = external_thread_control;
+  if (thread_control == NULL) {
+    thread_control = thread_control_create();
+    thread_control_set_status(thread_control, THREAD_CONTROL_STATUS_STARTED);
+  }
+  EndgameArgs endgame_args = {0};
```
jvc56, line 338: Now that we have to fill the EndgameArgs in more than one place, we should make an endgame_args_fill function that's called here and in config_fill_endgame_args so that the compiler can catch missing fields.

src/impl/play_chooser.c
```diff
@@ -0,0 +1,1076 @@
...
+                                 bool greedy_only, Move *out_move,
+                                 double *out_value, ErrorStack *error_stack) {
+  const int bag_letters = bag_get_letters(game_get_bag(game));
+  if (bag_letters < PEG_MIN_BAG || bag_letters > PEG_MAX_BAG) {
+    return false;
+  }
+  const PlayChooserStrategy *strategy = &play_chooser->strategy;
+  ThreadControl *thread_control = thread_control_create();
+  thread_control_set_status(thread_control, THREAD_CONTROL_STATUS_STARTED);
+  PegArgs peg_args = {0};
```
jvc56, line 465: PegArgs should have it's own peg_args_fill function.

jvc56 (changes requested):

src/ent/sim_results.c
olaugh: Done in c4c29edb — you're right on both counts, and the redundancy was more complete than the comment suggested.

`compare_simmed_plays` (post-#584) ranks by mean blended utility whenever `utility_w_spread > 0`, and the weight is stamped onto the **real** `SimmedPlay`s by `sim_results_set_utility_w_spread`, not just the display copies — so `sim_results_get_best_move_index`, which iterates the real plays, already ranks by utility. And BAI's `astar_index` is just argmax of the arm means (bai.h:344-351), which are means of the same per-sample utilities `utility_stat` accumulates. So `bai_result_get_best_arm` and the comparator were computing the same arm; the only divergences were tie-breaking (BAI takes the lowest index, the comparator applies the pass tiebreak) and the `< 0` not-yet-set fallback that existed only because of the BAI path.

Both functions now just call `sim_results_get_best_move_index`. I dropped the outdated claim and kept the still-true half of the comment — why a *zero* spread weight wants the win% comparator rather than BAI's arbitrary choice among saturated/tied arms — moving it onto `get_best_move_index`.

`sim`, `playchooser`, `peg`, `endgame` green at -O3 and -O0/ASAN (including sim_test.c's SKEIN case, which pins best-move selection under a nonzero spread weight).

src/impl/play_chooser.c
olaugh: Done in c4c29edb. `run_sim` now calls `sim_args_fill` instead of assigning `SimArgs` field by field (see the reply on the `utility_*` lines below for the parameters I had to add to make that possible).

One behavior note while converting: the old code set `bai_options.delta = 1.0` explicitly, but `sim_args_fill` only assigns `delta` on the `else` branch, leaving it untouched when the threshold is `NONE` — which is what the chooser passes. `delta` is only read under `BAI_THRESHOLD_GK16` (bai.h:270), so it was inert either way, but it meant `config_simulate`'s uninitialized `SimArgs args;` could leave the field holding stack garbage. I added `delta = 1.0` to the `NONE` branch so the fill leaves no field unset.

src/impl/play_chooser.c
olaugh: Done in c4c29edb. `sim_args_fill` now takes `utility_w_winpct`, `utility_w_spread` and `utility_spread_scale` as parameters instead of hardcoding the pure-win% defaults, so `play_chooser` passes its own weights through the fill.

This also removed the poke-after-fill pattern at the three existing call sites — `config_fill_sim_args` and both `config_fill_autoplay_args` fills were each assigning the same three fields immediately after calling it.

src/impl/play_chooser.c
olaugh: Done in c4c29edb. `endgame_args_fill` is in endgame.h, called from both `config_fill_endgame_args` and `play_chooser_run_endgame`.

It takes a parameter for **every** `EndgameArgs` field, which is what makes the compiler catch an added field at each call site — so it is 30 parameters, and `play_chooser` passes ~14 explicit `NULL`/`false`/`0`s for the callbacks, `actual_move`, `external_deadline_ns` and friends. Flagging that cost explicitly in case you'd rather trade the guarantee for a shorter signature; I went with the full version since the guarantee was the point of the request.

Both call sites previously relied on `memset`/`= {0}` for the fields they didn't care about, so every one of those defaults is now passed explicitly and behavior is unchanged (`DUAL_LEXICON_MODE_IGNORANT` is the zero value, etc.). The `memset` in `config_fill_endgame_args` is gone since the fill now covers the struct.

Two things I did **not** convert, since you scoped this to the two sites: peg.c's internal `EndgameArgs` literals and the ~20 in the endgame/benchmark tests. Both build literals with designated initializers and so opt out of the compiler check — noted in the function comment. Happy to convert peg.c's two if you want the check there too.

cppcheck flagged two `Game *` params as const-able once play_chooser started passing `game` into the const-taking fill; const-ified those (plus one knock-on `resolve_game`).

src/impl/play_chooser.c
olaugh: Done in c4c29edb. `peg_args_fill` is in peg.h, called from `config_fill_peg_args` and `play_chooser_run_peg`, and follows the same full-parameter shape as `endgame_args_fill` (30 params — same tradeoff noted on that thread).

As with the endgame args, every default the two call sites previously got from `memset`/`= {0}` is now explicit, so behavior is unchanged. `config_peg` still installs `poll` and the only/protect move sets after the call — the fill passes `NULL` for those, matching what the `memset` produced; I left that as-is rather than widen `config_fill_peg_args`'s contract, since `config_fill_analyze_args` shares it and wants them unset.

The peg tests keep their `PegArgs` literals and opt out of the check, same as the endgame tests.

src/impl/play_chooser.c
jvc56: oh, I missed the internal peg.c's EndgameArgs literals. It would be good to wrap those in a call to endgame_args_fill too but that can be done in a later PR.

jvc56 (approved):
LGTM. We can add more endgame arg fill function wrappers later

PR #594, opened 2026-07-15
Birdtest Upgrades

jvc56:
## Summary

Three mostly-independent additions to support distributed leavegen (birdtest) and general diagnostics:

- **`version` command** — prints the executable version (currently hardcoded `0.0.0`). Available via the CLI command and the `str_api_version` API entry point.

- **Forced rack list for leavegen** — `leavegen` takes an optional third argument, a file of racks (one per line), that restricts which racks its `RackList` ever selects as "rare." This lets a coordinator hand workers a fixed list of rare racks to evaluate instead of each worker maintaining its own independently-drifting rare-rack selection. Implementation went through two iterations before landing:
  - Initially added as a standalone `forceracksfile` mode on `autoplay`, with its own `ForceRackList` type and an `AUTOPLAY_RECORDER_TYPE_RACK_EQUITY` recorder.
  - Folded into `leavegen`'s existing `RackList` machinery instead, removing the standalone `ForceRackList` type and the separate recorder — `rack_list_create` now takes the optional forced-racks filename directly, and all racks start at count 0 with only the forced ones placed in the rare partition (re-applied on every `rack_list_reset`).

- **`-writerackequitycsv` config flag** — explicit, user-settable boolean (default `false`) that makes each leavegen generation write a `<rack>,<count>,<mean>` CSV via `rack_list_write_rack_equity_csv`, alongside the existing per-generation KLV/leaves/report files. Writes every observed rack, not just forced ones, so it's meaningful independent of a `forceracksfile` restriction (previously this was inferred implicitly from whether a `forceracksfile` was given).

## Other cleanup folded in

- `rack_list.c`: replaced direct `fgets`/`fprintf` with `getline_ignore_carriage_return`/`write_to_stream` for reading the forced-racks file and writing the equity CSV (same swap applied to pre-existing `fgets` loops in `klv_csv.c` and `autoplay_results.c`); moved a file-scoped enum constant into the top-of-file anonymous `enum` block; removed forward declarations by reordering functions ahead of their callers.
- Direct `#include <stdio.h>` for `FILE` in `rack_list.c` (was only reaching it transitively), to satisfy clang-tidy's `misc-include-cleaner` on CI's compiler.


olaugh:
## Code review (Claude, standing in for the errored-out Copilot pass)

Overall this is a clean, careful PR — the new error paths in `autoplay()` / `autoplay_shared_data_create` / `leavegen_shared_data_create` all release what they allocated (verified no leak on the `rack_list_create` failure path), the error-status coverage in the test is thorough, and the `fgets`→`getline` swaps correctly add the matching `free(line)`. A few issues, most-severe first.

### 1. Duplicate racks in a forceracksfile corrupt the rare partition — and can write out of bounds `[correctness]`
`src/impl/rack_list.c:265` (`rack_list_read_forced_rack_indices`) appends an index per line with no dedup, and `rack_list_restrict_to_forced_racks` (`rack_list.c:233`) does `partition_index++` once per entry:

```c
for (int i = 0; i < rack_list->num_forced_racks; i++) {
  const RackListItem *item =
      rack_list->racks_ordered_by_index[rack_list->forced_rack_indices[i]];
  partition_index++;
  rack_list_swap_items(rack_list, item->count_index, partition_index);
}
rack_list->partition_index = partition_index;
```

If the file lists the same rack twice, the second pass looks the item up again (now already at the front), swaps it back one slot, and pulls a **non-forced** rack into the rare partition — breaking the core "only forced racks are ever drawn" invariant silently.

Worse: `num_forced_racks` counts duplicates, so if it exceeds `number_of_racks`, `partition_index` runs past the end of `racks_partitioned_by_target_count` → **out-of-bounds write**. That's trivially reachable with the test's own `english_ab` distribution (only 8 full racks): a 9-line file of repeats overflows the array. ASAN would catch it; a release build would corrupt the heap.

**Fix:** dedup `forced_rack_indices` after reading (or reject a duplicate with `ERROR_STATUS_AUTOPLAY_FORCE_RACKS_MALFORMED_RACK`). Worth a test with a duplicate line.

### 2. "forced_racks" naming/comments for the equity CSV are stale after the redesign `[docs/naming]`
The CSV is now gated on `-writerackequitycsv` (independent of any forceracksfile) and writes **every observed rack**, per the function's own doc — but three places still describe the old behavior:
- `src/impl/autoplay.c:276` names the file `"%s_forced_racks.csv"`, and the comment at `:270` says *"When a forceracksfile was given, also dump…"* — actual guard is `write_rack_equity_csv`.
- `src/impl/config.c:1199-1205` (leavegen help) claims each generation *"also writes a `*_gen_<n>_forced_racks.csv` … for just those racks"* — neither gated on the forceracksfile nor limited to those racks.

Suggest renaming the suffix to something like `_rack_equity.csv` and correcting both comment and help text so the file name matches what it contains.

### 3. Dead error status `[cleanup]`
`ERROR_STATUS_AUTOPLAY_FORCE_RACKS_REQUIRES_RACK_EQUITY` (`src/util/io_util.h:82`) is declared but never referenced anywhere — a leftover from the earlier iteration where CSV-writing was inferred from the forceracksfile. Remove it.

### 4. Unbalanced parenthesis in the leavegen help text `[docs]`
`src/impl/config.c:1200` opens `"… ever forced as rare (used to run a distributed leavegen worker …"` and the `(` is never closed before the string ends at `:1205`. Minor, but it's user-visible help output.

### 5. Behavior change worth calling out for validation `[behavior]`
`src/impl/autoplay.c:700` now skips `rack_list_add_rack` when the forced move is a pass:

```c
if (move_get_type(forced_move) != GAME_EVENT_PASS) {
  rack_list_add_rack(...);
}
```

This is correct (a pass's equity is a sentinel that would poison the leave mean), but it also changes **normal, unrestricted** leavegen — previously such a rack recorded the sentinel. That's arguably a latent bug fix, but it's a silent change to existing output, so it'd be good to confirm via a game-pair / leavegen comparison that unrestricted runs are unchanged (or intentionally improved), not just the new forced path.

### Nits
- `src/impl/rack_list.c` has no trailing newline at EOF (the closing `}`); the `.h` was fixed to add one — do the same for the `.c`.
- New file-local helpers `rack_list_restrict_to_forced_racks` and `rack_list_read_forced_rack_indices` are non-`static` but not declared in the header (same as the pre-existing `rack_list_swap_items`). Could be `static` since they're only used within the file.
- No test covers the `-writerackequitycsv` output path or the duplicate-rack case (see #1).

Nothing here blocks the concept — #1 is the only correctness bug and it's a small fix (dedup on read).

*— Reviewed by Claude via Claude Code.*

olaugh (approved):

PR #602, opened 2026-07-21
Add embedding examples for the cmd_api shared library

martindemello:
Last of four stacked PRs for the `cmd_api` embedding API. **Based on #601** — review that one first.

Adds worked examples of embedding magpie and documents the API surface.

- `examples/generate_moves.c`: minimal C client, built against `bin/libmagpie.so` by the new `make examples` target (which sets an rpath so the binary finds the library in `bin/`).
- `examples/magpie.py`: Python ctypes wrapper with declared `argtypes`/`restype` for every API function, string ownership handled via `magpie_free_string`, plus an interactive REPL and sync/async helpers.
- README "API Library" section rewritten to document the embedding surface (`make libmagpie`, the command/output model, sync vs async, string ownership) and point at the examples.

Replaces the previous ad-hoc scratch files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ covhtml
 *.info
 *.out
 magpie*
+!examples/magpie.py
 dev/*.txt
 perf*
 .env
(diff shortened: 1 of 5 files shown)


jvc56 (approved):
LGTM! You can merge anytime.

PR #604, opened 2026-07-21
Rack blank alias, quit aliases, and a one-turn shmoves buffer

truthmane:
## Summary
- Accept "." as an alias for "?" (blank tile) in rack strings, since "." is a more natural/typo-resistant choice on some keyboards. Normalized in `rack_set_to_string`/`rack_set_to_string_unblanked` before parsing; board/move strings are unaffected since they already use "." for the unrelated played-through marker.
- Alias "exit" and "q" to the "quit" command in the interactive command loop.
- Let "shmoves" show the results of the last gen/sim for one more command after they'd otherwise be wiped (e.g. right after a commit), via a one-shot cached rendering captured just before the live move_list/sim_results are invalidated. The buffered view is display-only and never used as a basis for a new sim/commit.

## Test plan
- [x] `make magpie BUILD=release` and `make magpie_test BUILD=release` build clean
- [x] Full test suite (config, rack, command, sim, movegen, gameplay, gcg, autoplay, wmp, and more) passes under the ASAN/UBSAN dev build
- [x] `./cppcheck.sh`, `./tidy.sh`, and `python3 format.py --write` are all clean (pre-existing warnings unrelated, confirmed against unmodified `main` via `git stash`)
- [x] Manually verified all three features end-to-end via the CLI


jvc56:
The sim results from the previous turn should be saved as a SimResults instead of a string so that the users can actually interact with those results through the 'shmoves' command (like showing a specific move index or filtering).

truthmane:
Good catch — addressed in 8a287b6d. The buffered move_list is now a real duplicated `MoveList` object, so ordinary `shmoves` filter/index args (max count, coordinate, prefix) work against it after a commit, not just a frozen string.

For `SimResults` specifically, I couldn't safely do the same: it holds mutexes, per-ply heat maps, stats, and a `BAIResult`, with no existing duplicate function, and my first attempt at handing off the live pointer (and swapping in a fresh `SimResults` for `config->sim_results`) broke code that caches `config_get_sim_results()`'s pointer across commands expecting it to stay stable for the config's lifetime — caught it via an ASAN use-after-free while re-running the test suite. So the buffered sim view stays a fixed rendering captured at invalidation time. Happy to follow up with a proper `sim_results_duplicate()` in a separate PR if that's wanted, given the size/risk of getting that right.

jvc56:
Create a `sim_results_duplicate()` for this PR. We want a full deep copy duplicate so that when the game history is navigated to the previous turn, the user can perform all of the usual actions on the duplicated SimResults just as if they had just finished a sim for that turn.

truthmane:
Done in d031190. Added \`sim_results_duplicate()\` (plus \`bai_result_duplicate()\` and a \`simmed_play_duplicate()\` helper it needs) as a full, independent deep copy — every \`SimmedPlay\` with its stats and heat maps, the BAI result, and both racks. The buffered sim view now supports the same \`shmoves\` filter/index args and re-sorts as a sim that just finished, instead of a fixed rendering.

As before, each \`SimmedPlay\`'s PRNG and all the mutexes are freshly created rather than copied (the duplicate is never used to continue an active simulation) — matching the contract the existing \`simmed_play_copy()\` already established for its own partial copy.

Re-ran the full test suite (config, rack, sim, movegen, gameplay, autoplay, gcg, and more) under ASAN/UBSAN — all green, including new tests exercising filtering against both the buffered move_list and the buffered SimResults.

jvc56:
This is closer to what we want, but looking at the implementation, aligning the buffered result with the game state is tricky and might create some footguns. Instead of creating a single buffered_* result for the move list and sim results in the config, add a MoveList, SimResults, InferenceResults, and EndgameResults to the GameEvent struct. When any of those results are finished computing in config.c, a duplicate should be saved to their respective GameEvent in the GameHistory if a GameHistory is present (this would overwrite any existing results in the GameEvent with the more recent result for that GameEvent). Create the necessary duplicate functions for any results that might be missing a duplicate function. By adding results to the GameEvent, we can save all of the results for all game events, not just the previous game event, and it's easier to align the results with the correct game state. When showing a result, the config show prefer to show it's current result if present and valid, otherwise it should see if the current game event has that result, and then otherwise it should show the usual error that no results are present.

jvc56:
This looks good, it's pretty much all there. A few minor changes:

- Do not forward declare EndgameResults in src/ent/game_history.h, always prefer to use an include statement to make the struct available.
- The alias_method of the inference_results_duplicate does not need to be duplicated. The alias method is only used when running a sim. These duplicates will be used for display purposes. When a sim needs an inference result, it always reruns the inference, so the duplicate inference result in the game event does not need the alias_method
- Depending on the configuration, a sim will run an inference first and use that inference result in the sim calculation. For these cases, both the inference result and the sim result can be saved to the game event. So if the user runs a sim which uses inference, that game event should have an inference result and a sim result.
- In test_config_game_event_results_follow_navigation, add cases where sim results are successfully shown after a 'goto' to a previous move event more than 2 turns away and then after a 'goto' to a future move event more than 2 turns away. Set up the sim to be extremely quick (very few moves and iterations). We are not testing anything about the sim itself, just that it is available for display.

When you're finished, update the PR description.

jvc56 (approved):
LGTM. Two minor changes:

 - Delete the unused alias_method_duplicate function.
 - There are several places in the code where we use the human readable blank charactor '?". Instead of using the plain '?' we should define a enum in src/def/letter_distribution_defs.h (with the other existing enum constant values) to be '?' and use that constant instead of raw '?'

Once that's done, ask for a review from @olaugh . When he approves you can merge this in.

PR #606, opened 2026-07-22
Add timed PlayChooser players to autoplay

olaugh:
## Summary

- let either autoplay player use PlayChooser, with an independent total per-game clock specified in milliseconds
- have PlayChooser automatically use sim, pre-endgame, and endgame evaluation based on the position
- assess configurable started-period overtime penalties after each game and include them in final scores
- report PlayChooser clock use, overtime, and penalty points only when at least one player is active

## Interface

- `-pc1 <milliseconds>` / `-pc2 <milliseconds>`: `-1` disables PlayChooser for that player (the default), `0` is untimed, and a positive value sets the total game clock
- `-otpenalty <points>`: points deducted per started overtime period (default `10`)
- `-otperiod <milliseconds>`: overtime period length (default `60000`)

For blitz, `-otpenalty 1 -otperiod 1000` gives one point per started second.

## Autoplay results

Static equity played PlayChooser in paired CSW21 games with seed 606 and 10 workers. The default overtime rule was used.

| Run | Static wins | PlayChooser wins | Static mean score | PlayChooser mean score | Mean PC time/game | Overtime | Wall time |
|---|---:|---:|---:|---:|---:|---:|---:|
| 100 pairs, 1 s clock, concurrent games | 100 (50%) | 100 (50%) | 455.44 | 446.10 | 7.80 s | 188/200 games | 198.5 s |
| 10 pairs, 30 s clock, concurrent games | 7 (35%) | 13 (65%) | 449.00 | 460.45 | 27.29 s | 0/20 games | 56.2 s |
| 10 pairs, 30 s clock, 10 threads/game | 6 (30%) | 14 (70%) | 440.60 | 477.50 | 27.01 s | 1/20 games by 2.13 s | 542.2 s |

The 1-second run incurred 1,960 total penalty points (9.8/game). Before penalties, PlayChooser's estimated mean score was 455.90, essentially even with static's 455.44. This exposes weak deadline enforcement at blitz-scale clocks; follow-up is tracked in #607.

The 30-second samples are too small for a firm strength conclusion, but both favored PlayChooser. Giving all 10 threads to each game increased the observed margin from +11.45 to +36.90 points while making the run roughly 10 times slower wall-clock than concurrent games. All pairs in all three runs diverged.

## Verification

- release build
- strict `-Werror` address/undefined-behavior sanitizer build
- autoplay suite, including real player-1-only and both-player PlayChooser games plus player-2-only and neither-player reporting coverage
- PlayChooser suite
- parser bounds/error coverage for the new settings

### DIFF
--- a/src/impl/autoplay.c
+++ b/src/impl/autoplay.c
@@ -5,6 +5,7 @@
 #include "../compat/ctime.h"
 #include "../def/autoplay_defs.h"
 #include "../def/cpthread_defs.h"
+#include "../def/equity_defs.h"
 #include "../def/game_history_defs.h"
 #include "../def/letter_distribution_defs.h"
 #include "../def/players_data_defs.h"
--- a/src/impl/autoplay.c
+++ b/src/impl/autoplay.c
@@ -16,6 +16,7 @@
 #include "../ent/data_filepaths.h"
 #include "../ent/equity.h"
 #include "../ent/game.h"
+#include "../ent/game_timer.h"
 #include "../ent/inference_args.h"
 #include "../ent/inference_results.h"
 #include "../ent/klv.h"
@@ -35,9 +36,10 @@
 #include "../util/io_util.h"
 #include "../util/string_util.h"
 #include "gameplay.h"
+#include "play_chooser.h"
 #include "rack_list.h"
 #include "simmer.h"
-#include <limits.h>
+#include <math.h>
 #include <stdatomic.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -64,6 +66,21 @@ void autoplay_reset_total_sim_iterations(void) {
 // the game trajectory independent of sim throughput, so different RIT/BAI
 // variants run through the same sequence of positions.
 static _Atomic bool autoplay_bench_static_move_enabled;
+
+int autoplay_overtime_penalty_points(double overtime_seconds,
+                                     int points_per_period,
+                                     double period_seconds) {
+  if (overtime_seconds <= 0.0 || points_per_period <= 0 ||
+      period_seconds <= 0.0) {
+    return 0;
+  }
+  const double periods = ceil(overtime_seconds / period_seconds);
+  const int max_penalty_points = (int)EQUITY_MAX_DOUBLE;
+  if (periods >= (double)max_penalty_points / (double)points_per_period) {
+    return max_penalty_points;
+  }
+  return (int)periods * points_per_period;
+}
 
 void autoplay_set_bench_static_move(bool enabled) {
   atomic_store_explicit(&autoplay_bench_static_move_enabled, enabled,
@@ -356,12 +373,21 @@ AutoplayWorker *autoplay_worker_create(const AutoplayArgs *args,
   autoplay_worker->inference_results = NULL;
   autoplay_worker->error_stack = NULL;
 
-  // Only allocate sim structs if at least one of the players running a sim.
-  if (ap_args->p1_sim_args.num_plies > 0 ||
-      ap_args->p2_sim_args.num_plies > 0) {
+  const bool any_player_sims =
+      ap_args->p1_sim_args.num_plies > 0 || ap_args->p2_sim_args.num_plies > 0;
+  const bool any_player_uses_play_chooser =
+      ap_args->use_play_chooser[0] || ap_args->use_play_chooser[1];
+  // PlayChooser needs an error stack. The remaining objects are specific to
+  // the legacy autoplay simmer.
+  if (any_player_uses_play_chooser) {
+    autoplay_worker->error_stack = error_stack_create();
+  }
+  if (any_player_sims) {
     autoplay_worker->sim_results = sim_results_create(ap_args->cutoff);
     autoplay_worker->inference_results = inference_results_create(NULL);
-    autoplay_worker->error_stack = error_stack_create();
+    if (autoplay_worker->error_stack == NULL) {
+      autoplay_worker->error_stack = error_stack_create();
+    }
     rack_set_dist_size_and_reset(&autoplay_worker->target_played_tiles,
                                  ld_get_size(ap_args->game_args->ld));
     rack_set_dist_size_and_reset(&autoplay_worker->nontarget_known_rack,
@@ -493,8 +519,19 @@ typedef struct GameRunner {
   // inference
   Game *game_one_move_behind;
   Move previous_move;
+  Move play_chooser_move;
+  PlayChooser *play_choosers[2];
+  GameTimer game_timer;
+  AutoplayGameTiming timing;
   AutoplaySharedData *shared_data;
 } GameRunner;
+
+static void game_runner_destroy_play_choosers(GameRunner *game_runner) {
+  for (int player_index = 0; player_index < 2; player_index++) {
+    play_chooser_destroy(game_runner->play_choosers[player_index]);
+    game_runner->play_choosers[player_index] = NULL;
+  }
+}
 
 GameRunner *game_runner_create(AutoplayWorker *autoplay_worker) {
   const AutoplayArgs *args = &autoplay_worker->args;
@@ -507,6 +544,10 @@ GameRunner *game_runner_create(AutoplayWorker *autoplay_worker) {
   }
   game_runner->pair_game_number =
       0; // Will be set in game_runner_start if using pairs
+  game_runner->play_choosers[0] = NULL;
+  game_runner->play_choosers[1] = NULL;
+  game_timer_reset(&game_runner->game_timer, 0.0);
+  game_runner->timing = (AutoplayGameTiming){0};
   return game_runner;
 }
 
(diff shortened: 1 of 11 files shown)


olaugh (commented):
Reviewed the timed-PlayChooser autoplay feature. The shape is clean: per-player ms clock, auto-selection of sim/PEG/endgame by position, per-started-period overtime penalties assessed at game end and folded into final scores, and reporting gated on at least one active player. The penalty math (`autoplay_overtime_penalty_points`) is tidy — `ceil(overtime/period) * points_per_period` with the zero-guards and the `EQUITY_MAX_DOUBLE` overflow cap — and it's applied via `player_add_to_score` before the result is recorded, so it correctly flows into win/loss/spread. One meaningful caveat inline, plus nice to see the test coverage across autoplay/config/play_chooser.

src/impl/autoplay.c
```diff
@@ -772,6 +855,28 @@ const Move *game_runner_play_move(AutoplayWorker *autoplay_worker,
...
+    game_runner->timing.seconds_used[player_index] =
+        game_timer_get_seconds_used(&game_runner->game_timer, player_index);
+    game_runner->timing.overtime_seconds[player_index] =
+        game_timer_get_overtime_seconds(&game_runner->game_timer, player_index);
+    const int penalty_points = autoplay_overtime_penalty_points(
+        game_runner->timing.overtime_seconds[player_index],
+        args->overtime_penalty_points, args->overtime_period_seconds);
+    game_runner->timing.penalty_points[player_index] = penalty_points;
+    player_add_to_score(game_get_player(game_runner->game, player_index),
+                        -int_to_equity(penalty_points));
```
olaugh, line 876: Worth calling out (in the README/`-pc` help, and maybe a note here): once a player is on a wall-clock budget, autoplay results stop being **reproducible from the seed**. Move choice now depends on how much search fits in the budget, which varies with machine load, thread count, and build — so `-pc1 5000 -seed 42` won't replay identically, and the deducted penalty itself is nondeterministic. That also partly undercuts the premise of game pairs (`-gp`), whose whole point is deterministic variance reduction over identical draws: with timing on, the two paired games are no longer a clean controlled comparison. Not a bug — it's inherent to timed play — but since the rest of autoplay is seed-deterministic, users will assume these runs are too. A one-line warning when `-pc` is combined with `-gp` (or just in the help text) would set expectations. Are the timing-independent aggregates (win%, penalty points) the intended output to read here, with the per-game reproducibility explicitly not guaranteed?

olaugh: Yes — the timing-independent aggregates (win%, spread, penalty points) are the intended output, and non-deterministic per-game results are expected and acceptable here. Documented rather than changed, in bae89321.

Three places, since someone can arrive at this from any of them:

- **README**, alongside the `-pc1`/`-pc2` examples.
- **`-pc1`/`-pc2` help text** — where someone enabling a clock actually looks.
- **`-gp` help text**, which is the part that was actively misleading. It asserted without qualification that "the games are deterministic for a given starting seed" and that identical decisions produce identical games. That stops being true with a clock, so it now says so and notes that pairs lose their variance-reduction property, since the two games can diverge for reasons unrelated to strategy.

Documentation only, no behavior change.

olaugh:
Sample output from a very short run, plus a demonstration of the non-reproducibility the docs now describe.

## Sample output

Two games, a deliberately tiny 500 ms clock so it finishes quickly, single-threaded:

```
magpie> set -lex CSW21 -threads 1 -hr true -savesettings false
magpie> autoplay games 2 -pc1 500 -pc2 500 -seed 42
```

```
All Games

Games Played: 2
Turns Played: 45
Turns per Game: 22.50 0.71

Game End Reason None:      0 (0.00%)
Game End Reason Standard:  2 (100.00%)
Game End Reason Pass:      0 (0.00%)

                         Player 1                 Player 2
Total:                   2.0 (100.00%)            0.0 (0.00%)
Wins:                    2 (100.00%)              0 (0.00%)
Losses:                  0 (0.00%)                2 (100.00%)
Ties:                    0 (0.00%)                0 (0.00%)
Score:                   507.00 22.63             313.00 2.83

Player 1 is better with confidence: 76.024994%

PlayChooser Timing

                         Player 1                 Player 2
Time control (ms):       500                      500
Mean time used (ms):     818.607                  4419.945
Total overtime (ms):     637.215                  7839.891
Overtime games:          2                        2
Penalty points:          20                       20
Rate: 10 points per started 60000 ms
```

The `PlayChooser Timing` block is the new section, appended when at least one player is active. A 500 ms budget is far below what a turn actually needs, so both players sit in overtime the whole game and take the full 20-point deduction — which is what makes it fast enough to paste here.

## The reproducibility point, demonstrated

Same command, same `-seed 42`, three consecutive runs:

| run | turns played | P1 score | P2 score | P1 mean time (ms) | P2 mean time (ms) |
|---|---|---|---|---|---|
| 1 | 45 | 507.00 | 313.00 | 818.607 | 4419.945 |
| 2 | 51 | 422.00 | 367.00 | 2738.503 | 730.087 |
| 3 | 48 | 460.50 | 346.50 | 2036.649 | 4339.918 |

Different game lengths, different scores, and mean time used swinging by 5x between runs — including which player burns the most clock flipping between runs 1 and 2.

The control confirms this is the clock and not something else. With PlayChooser off (`-pc1 -1 -pc2 -1`), same seed, two runs are **byte-identical** and no timing block is emitted:

```
$ diff ctl1.txt ctl2.txt && echo IDENTICAL
IDENTICAL
```

So seed-determinism is intact for ordinary autoplay and is lost exactly when a clock is introduced, which is what the README and the `-pc1`/`-pc2` and `-gp` help text now say.

One practical note for anyone reproducing this: settings persist between sessions, so `-savesettings false` matters here. Without it a clock set in an earlier session carries into what looks like an unclocked run — I hit exactly that while putting these numbers together, and it made a clean control look nondeterministic.

Environment: release build (`-O3 -flto -march=native`), macOS arm64, `-threads 1`, CSW21.

olaugh (commented):
Self-review before handing this over. I traced the concurrency, the penalty arithmetic, and the config wiring against current `main`, and built and ran the branch. Recording what I verified so the next reviewer does not have to re-derive it, then four things worth a look — none of which I think block merge.

## Verified

**Penalty ordering is right.** `game_runner_assess_overtime` runs after the game loop and before `autoplay_add_game`, so `player_add_to_score(-int_to_equity(penalty))` lands before `game_data_add_game` reads the scores. The deduction therefore flows into win/loss/tie, the score stat, and spread, which is the intent.

**The overflow guard holds.** `autoplay_overtime_penalty_points` early-returns on any non-positive input, so the division is safe. The `periods >= max_penalty_points / points_per_period` test means the surviving multiply cannot exceed `EQUITY_MAX_DOUBLE`, and `(int)periods` is itself bounded by that same value, so neither the cast nor the product overflows. `EQUITY_MAX_DOUBLE` is `EQUITY_MAX_VALUE / EQUITY_RESOLUTION`, so the capped `-int_to_equity(...)` stays inside `int32_t`.

**No race on the new recorder state.** `autoplay_results_create_empty_copy` shares `recorder_context` by pointer (`owns_recorder_context = false`), and `autoplay_results_set_play_chooser_config` is called at autoplay.c:1131, before workers are created at :1201. So it is write-once before any thread starts, every worker observes it, and the consolidated primary prints from the same struct. Consolidation sums all four per-player accumulators; nothing is dropped.

**Game pairs do not contaminate each other's clocks.** Each `GameRunner` owns its `GameTimer`, and `game_timer_start_turn` / `end_turn` only touch that game's clock, so interleaving the two games in one worker thread is safe. `game_timer_end_turn` early-returns when `ticking_player_index < 0`, so the extra call at the top of `assess_overtime` is a genuine no-op rather than a double count.

**Seed derivation is fine.** `strategy.seed = seed + player_index` looked like it could collide with the next game's seed, but per-game seeds come from `prng_next(shared_data->prng)`, not a counter, so there is no systematic correlation. Worth stating explicitly because it would have been a real bug against sequential seeds.

**`RecorderArgs args = {0}` is a latent-bug fix, not noise.** `game_data_add_game` now reads `args->timing` unconditionally, and the previous uninitialized `RecorderArgs args;` would have handed it a garbage pointer on the paths that do not set it.

## Worth a look

**1. `-pc` is silently ignored for leavegen.** `use_play_chooser` is gated on `autoplay_type == AUTOPLAY_TYPE_DEFAULT`, so `autoplay leaves ... -pc1 5000` drops the clock with no error, no warning, and no timing block to hint that anything was discarded. Settings are session-sticky, so the likely path here is setting a clock for a normal run and then starting a leavegen run in the same session. An error, or at minimum a warning, seems better than silence.

**2. Effective clock depends on `-threads`.** Under the default inter-game parallelism, N games run concurrently, so a 30s budget buys roughly 1/N the compute it would at `-threads 1`. That is systematic rather than jitter: win% from a `-threads 1` run and a `-threads 8` run are not comparable even in aggregate, which is a stronger statement than the "not reproducible" caveat now in the docs. Is that the intended reading, and should the docs say it outright?

**3. `-pc1 -0.5` disables, but only `-1` is documented.** The accepted range is `[-1, 1e12]` and the disable test is `>= 0.0`, so every value in `[-1, 0)` disables. Either say "negative disables" in the help text or reject anything that is not exactly `-1`.

## Nits

- The printed final board (`-printboards`) is emitted after the penalty is applied, so its score is the post-deduction one. I checked, and it agrees with the reported `Score:` line — no inconsistency between them — but it will not reconcile with the sum of the plays shown turn by turn. Annotating the deduction in that block might save someone a confused subtraction.
- The new `%lu` conversions take `uint64_t`, which is not portable on LLP64. This matches the existing `%lu` for `gd->total_games` in the same function, so it is consistent with the surrounding code rather than something this PR introduces. Non-blocking.

Build clean at `-O3`, and `config` / `autoplay` / `playchooser` / `peg` all pass locally on top of current `main`.

olaugh:
Correcting point 2 of my review above, and documenting the parallelism in `a0e5bacf`.

**The correction.** I wrote that under the default threading a 30s budget "buys roughly 1/N the compute it would at `-threads 1`." That is wrong. `num_worker_threads_per_sim` starts at 1 and is only raised to `config->num_threads` under intra-game parallelism, so in the default per-game mode each concurrent game already gets its own thread and a single-threaded chooser. Per-game compute per wall-second is roughly constant in `-threads`, not divided by it.

**What actually happens:**

- **`-mtmode pgp` (default).** `-threads` games run concurrently, each chooser single-threaded. Raising `-threads` finishes the run sooner but gives no player more search per second of clock — until the thread count exceeds available cores, at which point games contend and the same budget buys measurably less.
- **`-mtmode igp`.** `num_threads` is forced to 1 and the full `-threads` count goes to the chooser, so one game runs at a time with the whole machine behind each move. The same clock buys roughly `-threads` times as much search per move.

So the comparability rule stands, but for a different reason than I gave: the large effect is `-mtmode`, and `-threads` only matters within `pgp` once you oversubscribe cores. Timed runs are comparable only against runs with the same `-mtmode` and `-threads`.

Documented in the README beside the reproducibility notes and in the `-pc1`/`-pc2` help text. Documentation only, no behavior change; `config` / `autoplay` / `playchooser` still pass.

jvc56 (approved):

PR #608, opened 2026-07-22
Bound PlayChooser overtime by penalty period

### LINKED ISSUE #607: Enforce PlayChooser deadlines tightly enough for blitz clocks
## Problem

PlayChooser's total game clock is currently a soft budget. At very short time controls, solver calls can overrun badly enough that overtime penalties dominate the result.

Observed while validating #606 with static equity versus PlayChooser, using CSW21, seed 606, 10 workers, and the default 10-points-per-started-minute penalty:

- 100 pairs / 200 games at a 1,000 ms PlayChooser game clock
- mean PlayChooser time: 7,804.7 ms/game
- overtime in 188/200 games (94%)
- 1,362.2 seconds total overtime
- 1,960 total penalty points, or 9.8 points/game

Before penalties, PlayChooser's estimated mean score was 455.90 versus static's 455.44. After penalties, it fell to 446.10, so clock overrun rather than playing strength drove the result.

The behavior is much better at a 30-second clock:

- concurrent-game run: 0/20 games overtime, 27.29 seconds mean time used
- 10-threads-per-game run: 1/20 games overtime by 2.13 seconds, 27.01 seconds mean time used

## Likely areas to investigate

- minimum per-move budget allocation under very short total clocks
- deadline propagation through simulation, pre-endgame, and endgame searches
- coarse or non-preemptive stopping inside solver calls
- reserving enough clock for later turns and unavoidable cleanup overhead

## Desired behavior

PlayChooser should respect short total game clocks closely enough that overtime is exceptional rather than normal. Add regression coverage for blitz-scale clocks and report the maximum/mean overrun so future changes cannot silently weaken deadline enforcement.

olaugh:
## Summary

Timed PlayChooser searches could overrun the game clock: a forced 50 ms floor kept launching searches after flag fall, direct endgames lacked an absolute deadline inside alpha-beta, and PEG could finish long in-flight playouts after its budget expired.

- Divide safely spendable clock time across estimated remaining plays and use static play when the budget is too small.
- In overtime, spend only the remainder of the current penalty period, reserving 1% of that period plus per-play overhead.
- Give direct endgames an absolute mid-search deadline and check PEG cancellation between scenario orderings and playout plies.
- Allocate PEG workers' endgame transposition tables only when an exact leaf needs them.
- Reject interrupted or incompletely scored live PEG candidates before publishing rankings or captured outcomes.
- Preserve explicitly configured fixed move budgets, including budgets below the automatic static-fallback thresholds.

Fixes #607. Based on current `main`, including the merged #606 PlayChooser integration and #609 outcome rollback fix.

### Interrupted PEG candidates must not count as completed

On the existing ONYX/OXY fixture, interrupting the second candidate after its first scenario could give that candidate full standing in the live ranking despite evaluating only **1 of 8 required scenario weights**. In the first-stage reproduction, the partial candidate even ranked above the fully scored candidate. The pooled path already checked weight completeness; the live path checked only elapsed time, so a user interrupt bypassed that protection.

Live acceptance now checks both cancellation and the candidate's complete weight from the preceding stage. It releases the partial outcome capture and uses the existing rollback if fewer than two candidates completed.

`test_peg_interrupted_candidate_outcomes` exercises the interaction with #609 deterministically: one worker completes the first 3-ply candidate, then a scenario callback interrupts the second. There is no wall-clock timeout in this test.

| | Before correction | After correction |
|---|---|---|
| Partially scored candidate | Counted as completed | Excluded |
| Accepted stage | Incomplete 3-ply stage | Previous completed 2-ply stage |
| Published candidates | Could contain weight 1 alongside weight 8 | Both retain weight 8 across 7 unique scenarios |
| Captured outcomes | Could publish partial/deeper rows | Both remain at 2-ply; captured weights each sum to 8 |

Run the focused regression with `./bin/magpie_test peginterrupt`.

### A fixed 125 ms budget must actually reach the endgame solver

The budget helper returned an explicit 0.125 seconds correctly, but move selection then applied the automatic 250 ms endgame minimum and silently chose static play. The minimum-budget fallback now applies only when the budget is not explicitly fixed.

`test_play_chooser_fixed_short_budget` uses a sparse board with `AT` at 8H, racks `AEQ/IZ`, a bag explicitly emptied by the test, and an expired game clock:

| Budget policy | Before correction | After correction |
|---|---|---|
| Clock-derived, no overtime allowance | Static `I6 QA(T)` for 13 | Same static fallback |
| Explicit 125 ms, two-ply endgame | Static `I6 QA(T)` for 13 | Search chooses `9I AE` for 6, denying the immediate `Z(AT)I` outplay |

This tests public move selection, rather than just asserting the value returned by the budget helper. Both regressions also run in the normal `peg` and `playchooser` suites.

Run the focused regression with `./bin/magpie_test playchooserfixed`.

## Verification

Both new regressions fail against the code before these corrections and pass with them.

```sh
python3 format.py --write
python3 find_circ_deps.py
make magpie_test BUILD=release
./bin/magpie_test playchooser peg
make magpie_test BUILD=dev BIN_DIR=bin/review608-dev
./bin/review608-dev/magpie_test peginterrupt playchooserfixed
```

- Release PlayChooser and PEG suites: passed.
- Focused ASAN/UBSAN regressions: passed.
- Formatting and circular dependency checks: passed.
- GitHub CI: see the checks on the updated branch for the full Linux/static-analysis/WASM matrix.

## Original July performance measurements

The measurements below are retained from the original implementation; they were not rerun for the two correctness corrections above.

A 20-second sample of the failing 1-second workload attributed about 73% of PlayChooser time to direct bag-empty endgame solving, 14% to simulation, 4% to PEG, and 8% to setup/static work. Direct endgame had only the solver hard limit, which is checked at depth boundaries, rather than the absolute deadline used inside alpha-beta. Once the clock reached zero, the old forced 50 ms floor also kept launching searches on every remaining move.

After adding strict clock budgeting, the residual flag falls were all at the 4-tile PEG stage: PEG received 60-109 ms budgets but returned 320-1,395 ms later under the representative 10 concurrent games x 10 threads workload. PEG checked the deadline at stage and candidate boundaries, but eagerly allocated an endgame TT for every worker before the greedy stage (which does not use them), and an in-flight scenario could complete every bag ordering and a full greedy playout without another deadline check. This PR makes those TTs lazy and propagates cancellation through that inner work rather than disabling PEG at short controls.

### Results

Same 100-pair / 200-game, 1-second-per-game command, seed 606, with ten concurrent games and ten threads per game:

| | Before | After |
|---|---:|---:|
| Mean PlayChooser time/game | 6,311.577 ms | 794.270 ms |
| Overtime games | 188/200 | 0/200 |
| Total overtime | 1,063,224.950 ms | 0 ms |
| Penalty points | 1,920 | 0 |

At the normal 30-second control, PlayChooser won 14/20 against static, used 27,044.906 ms/game on average, and had zero overtime games.


src/impl/peg.c
```diff
@@ -763,15 +820,28 @@ static int32_t peg_pessimistic_playout(Game *game, int mover_idx,
...
+        }
+        break;
+      }
       game_copy(branch, game);
       play_move(move_list_get_move(opp_ml, i), branch, NULL);
       const int32_t mover_net =
-          peg_greedy_playout(branch, mover_idx, rollout_ml);
+          peg_greedy_playout(branch, mover_idx, rollout_ml, deadline_ns,
+                             thread_control, interrupted);
+      if (interrupted != NULL && *interrupted) {
```
olaugh, line 834: The inner handling here is right — on interrupt you `break` before comparing the partial `mover_net`, so the half-finished greedy playout never influences `worst_idx`. But `peg_pessimistic_playout` still returns a `spread` computed from the partial game state after the outer break. The correctness of the whole timed-PEG path then depends on the scenario aggregator treating an interrupted playout as *not fully scored* — i.e. its weight must be excluded so the candidate's `weight_sum` falls short and gets cut by the same acceptance logic #609 just hardened. Could you confirm (or point to) where an interrupted playout's partial `spread` is kept out of the weighted score/`weight_sum`? If a partial spread ever lands in a scenario that still counts toward a full `weight_sum`, you'd get exactly the #609-class symptom (a candidate that looks complete with a bad score). This is the one spot I'd want a test for: interrupt mid-playout and assert the candidate is discarded, not ranked.

src/impl/peg.c
```diff
@@ -123,6 +124,16 @@ typedef struct PegWorker {
   PegNestFrame *nest_all;
 } PegWorker;
 
+// The greedy seed does not use an endgame transposition table. Allocate one
+// only when a worker actually reaches an exact leaf so a short-budget PEG does
+// not spend most of its clock zeroing tables it will never consult.
+static TranspositionTable *peg_worker_get_endgame_tt(PegWorker *worker) {
```
olaugh, line 130: Lazy per-worker TT alloc is race-free (each `PegWorker` owns its `eg_tt`), good. Two things to double-check: (1) teardown frees `eg_tt` only when non-NULL for workers that never reached an exact leaf, and (2) `eg_tt_fraction` is always initialized before the first `peg_worker_get_endgame_tt` call — since a zero/garbage fraction would size the table wrong on first lazy alloc rather than at worker setup where it'd be obvious. A one-liner asserting `eg_tt_fraction > 0` here would catch an uninitialized-field regression cheaply.

olaugh (commented):
Reviewed the overtime-bounding PR. The budget/reserve model reads well — the reserve-aware window (clock reserve + 1%-of-period overtime guard + per-play reserve) with a static fallback when there's no useful search time is a sound replacement for the flat 50ms floor, and the absolute mid-search endgame deadline plus interruptible PEG are the right levers for a hard per-move budget. Lazy endgame-TT allocation is a nice win for short budgets. Two things worth a look inline. Note this is closely coupled to #609: that PR fixed partial-stage results leaking, and this one *adds* partial-result surface (mid-playout / between-ordering interrupts), so the two need to compose.

src/impl/peg.c
olaugh: The greedy/pessimistic path passes `&ctx->interrupted` into the playout. In [`peg_eval_split`](https://github.com/jvc56/MAGPIE/blob/385961ddc13df36b5c07405abea779f326c07333/src/impl/peg.c#L1825-L1833), an interrupted split returns before adding its weight or spread to the aggregate. The candidate-worker path also marks interrupted candidates invalid with zero weight.

There was a downstream gap in live acceptance: a user interrupt could leave a candidate with weight 1 instead of 8, yet the live caller counted it as complete because it only checked the deadline. Commit 406c9a66 fixes that by [checking cancellation and full candidate weight before publishing](https://github.com/jvc56/MAGPIE/blob/385961ddc13df36b5c07405abea779f326c07333/src/impl/peg.c#L3014-L3025), including its outcome capture. This composes with the now-merged #609 rollback.

The new [`peginterrupt` regression](https://github.com/jvc56/MAGPIE/blob/385961ddc13df36b5c07405abea779f326c07333/test/peg_test.c#L1365-L1427) completes one 3-ply candidate and interrupts the next after its first scenario. It verifies rollback to the completed 2-ply ranking, both candidates' full weight of 8, and both captures remaining at 2-ply with total weight 8. It fails before the fix and passes in release and ASAN/UBSAN builds.

One coverage detail remains: this test interrupts **between scenarios**, not inside a greedy/pessimistic playout. Leaving this thread open for the specific mid-playout regression requested here; the current test should not be described as covering that exact interruption point.

src/impl/peg.c
olaugh: Confirmed both invariants, and added the suggested `assert(worker->eg_tt_fraction > 0.0)` at [first lazy allocation](https://github.com/jvc56/MAGPIE/blob/385961ddc13df36b5c07405abea779f326c07333/src/impl/peg.c#L132-L136) in 385961dd.

Every scratch slot, including the pool helper's slot, is [initialized before any evaluation jobs run](https://github.com/jvc56/MAGPIE/blob/385961ddc13df36b5c07405abea779f326c07333/src/impl/peg.c#L2739-L2759): `eg_tt = NULL` and `eg_tt_fraction = min(0.05, 0.25 / n_scratch)`, with `n_scratch >= 1`. Workers that never reach an exact leaf therefore retain NULL.

Teardown calls `transposition_table_destroy` for each slot; that function [returns immediately for NULL](https://github.com/jvc56/MAGPIE/blob/385961ddc13df36b5c07405abea779f326c07333/src/ent/transposition_table.h#L206-L213), before dereferencing or freeing the table. No additional conditional is needed at the caller.

The focused release tests passed, and `peginterrupt` passed under ASAN/UBSAN with the new assertion enabled, exercising lazy exact-leaf allocation and teardown. Resolving this thread.

jvc56 (approved):

PR #609, opened 2026-07-22
peg: discard outcomes from rolled-back stages

olaugh:
## Bug

When a time limit or interrupt lets fewer than two candidates finish a deeper PEG halving stage, the solver discards that stage and restores the preceding fidelity's ranking. The per-candidate outcome captures were published before that acceptance check, however, so a completed candidate could retain deeper outcomes beside shallower win/tie/loss counts.

This produces the contradiction from the report: the stage table shows `1/16` candidates completed at 3-ply, the graded row correctly falls back to 2-ply with `8/0/0`, but its outcomes cell comes from the discarded 3-ply evaluation and says `L: R T`.

## Fix

- Keep each halving stage's outcome captures in a stage-local array.
- Merge completed captures into the persistent outcome store only after at least two candidates finish and the stage is accepted.
- Free the stage-local captures when the stage is rolled back.
- Apply the same ownership flow to live sequential evaluation and non-live pooled evaluation.

## Regression test

The new test uses a callback to interrupt the second halving stage immediately after its first candidate finishes. This reproduces the failure deterministically without relying on a wall-clock race. It verifies that the accepted result and both candidates' outcome captures remain at 2-ply.

Before the fix, the first capture incorrectly reports fidelity 3. After the fix, both report fidelity 2.

## Exact report reproduction

Replayed the supplied `r2.gcg` at event 23 (`goto 23`) and ran:

```
peg -pegoutcomes true -tlim 15
```

Both revisions reached the same interrupted state: 302 greedy candidates, all 32 candidates complete at 2-ply, and 1/16 complete at the discarded 3-ply stage.

- Parent commit: `14B M(OS)` showed `8 wins, 0 ties, 0 losses, 100.0%` but `L: R T`.
- This PR: the same row shows `8 wins, 0 ties, 0 losses, 100.0%` and `always wins`.

The remaining accepted 2-ply rows retained their expected outcomes.

## Ownership invariant, documented

Review noted that the rollback's `peg_cand_outcomes_destroy_array` is the subtle safety hinge here, and that seeing why it is not a double free requires tracing all three stage exits — no single hunk shows it. That whole-picture invariant is now recorded at the call site:

- the completed prefix `[0, done_count)` reaching the rollback was never upserted, since accept is the only publisher;
- the cut tail was already freed by whichever evaluation path cut it (live frees the deadline-cut candidate on its `break`; batch frees `[done_count, eval_count)` after stable-partitioning);
- accept upserts exactly `[0, done_count)` and then frees only the array.

So every `rows` allocation is freed exactly once regardless of which exit runs.

## Also: name the outcome buckets

Folded in a small readability cleanup to the outcomes renderer, since it is the same subsystem and the same reviewer context.

`peg_string.c` classified every draw into win/loss/tie as a bare `int`, with the meaning carried only by comments — and one had already gone stale: `PegOutcomeTok.bucket` still read `0 = win, 1 = loss` after the tie bucket was added in #603. The unrolled cardinality sums (`any[0] + any[1] + any[2]`, and the matching `fam->seen` sum) would also silently undercount if a fourth bucket ever appeared.

`PEG_OUTCOME_BUCKET_{WIN,LOSS,TIE}` plus a trailing `NUMBER_OF_PEG_OUTCOME_BUCKETS` now live in a top-of-file anonymous enum, matching how `sim_string.c` and `bag_string.c` hold their file-private constants. They replace the literals in array extents, loop bounds, the classifier's returns, and the implied-bucket comparisons; the unrolled sums and the triple `seen[]` assignment become loops over the bucket range. The `labels` / `otherwise_word` lookup tables get designated initializers so the value-to-index binding is explicit — `imply_pref` and `display_order` stay positional, since they map preference and display rank to bucket rather than being bucket-indexed.

Pure refactor with no behavior change; the renderer tests assert exact output strings, so they pin that.

## Verification

- `python3 format.py --write`
- `make magpie_test BUILD=release`
- `./bin/magpie_test peg`, `./bin/magpie_test config`
- `make magpie_test` (dev/ASAN/UBSAN build with `-Werror`)
- `./cppcheck.sh` (2.17.1) clean, `python3 find_circ_deps.py` no cycles
- Exact `r2.gcg` parent-vs-fix reproduction above
- Complete `r2.gcg` 3-ply run: all 32/32 displayed candidates completed, with tie draws rendered
- All 17 CI checks green

## Note on the diff

#603 has since merged to `main`, and this branch is merged up to it, so its commits no longer appear in this diff. What remains is `peg.c` (the fix), `peg_test.c` (the regression test), and `peg_string.c` (the bucket enum).

### DIFF
--- a/src/impl/peg.c
+++ b/src/impl/peg.c
@@ -2799,6 +2799,14 @@ void peg_solve(const PegArgs *args, PegResult *out, ErrorStack *error_stack) {
       PegRankedCand *restaged =
           malloc_or_die((size_t)eval_count * sizeof(PegRankedCand));
       int done_count = eval_count;
+      // Keep this stage's captures separate until the stage is accepted. A
+      // stage with fewer than two completed candidates is rolled back to the
+      // preceding fidelity, so publishing its captures early would pair deeper
+      // outcomes with the restored shallower ranking.
+      PegCandOutcomes *stage_outcomes =
+          args->include_per_scenario
+              ? malloc_or_die((size_t)eval_count * sizeof(PegCandOutcomes))
+              : NULL;
       if (args->poll != NULL) {
         // Live mode: evaluate one candidate at a time so each completion
         // updates the pollable leaderboard (for `sta`/`shpeg`) and so we can
@@ -2827,13 +2835,12 @@ void peg_solve(const PegArgs *args, PegResult *out, ErrorStack *error_stack) {
           const int64_t cand_start_ns = ctimer_monotonic_ns();
           peg_poll_set_evaluating(args->poll, cand_idx, cand_start_ns);
           ctimer_start(&cand_timer);
-          PegCandOutcomes cand_oc;
           peg_eval_candidates_scenario(
               pool, workers, prepared_base, mover_idx, unseen, ld_size, ld,
               bag_size, &moves[cand_idx], 1, args->opp_model, args->inner_top_k,
               stage_fidelity, scenario_stride, deadline_ns,
               args->thread_control, &inner, /*poll=*/NULL, &restaged[cand_idx],
-              args->include_per_scenario ? &cand_oc : NULL);
+              args->include_per_scenario ? &stage_outcomes[cand_idx] : NULL);
           restaged[cand_idx].eval_seconds = ctimer_elapsed_seconds(&cand_timer);
           peg_poll_set_evaluating(args->poll, -1, 0);
           // If the deadline passed while this candidate was evaluating, some of
@@ -2841,14 +2848,11 @@ void peg_solve(const PegArgs *args, PegResult *out, ErrorStack *error_stack) {
           // rather than show or rank a partial result, and stop the stage.
           if (deadline_ns != 0 && ctimer_monotonic_ns() >= deadline_ns) {
             if (args->include_per_scenario) {
-              free(cand_oc.rows);
+              free(stage_outcomes[cand_idx].rows);
             }
             break;
           }
           done_count = cand_idx + 1;
-          if (args->include_per_scenario) {
-            peg_outcomes_store_upsert(&oc_store, &oc_n, &oc_cap, &cand_oc);
-          }
           // Surface this finished candidate into the leaderboard, then stream
           // the updated ranking to the caller right away — every candidate's
           // result prints as soon as it finishes, so the deep stages fill in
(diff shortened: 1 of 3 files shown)


src/impl/peg.c
```diff
@@ -2941,13 +2936,22 @@ void peg_solve(const PegArgs *args, PegResult *out, ErrorStack *error_stack) {
       if (done_count < 2) {
         n_graded = n_graded_before_stage;
         free(restaged);
+        peg_cand_outcomes_destroy_array(stage_outcomes, done_count);
```
olaugh, line 2947: This is the subtle safety hinge and it's worth a one-line note here for future readers: reaching this rollback means `stage_outcomes[0..done_count)` are the completed captures that were *never* upserted (the fix moved the upsert past this check), so destroying them is freeing owned memory, not double-freeing published rows. The symmetric invariant is that the cut tail `[done_count, eval_count)` was already freed in the batch/live evaluation above and the accept path below upserts exactly `[0, done_count)` — so every `rows` allocation is freed exactly once on all three exits. That whole-picture invariant isn't obvious from any single hunk; a comment tying the three exits together would save the next person the trace I just did.

olaugh (commented):
Traced through this one carefully — the fix is correct and the ownership is clean. Deferring the `peg_outcomes_store_upsert` until after the `done_count < 2` acceptance check is exactly the right place: a rolled-back stage now destroys its captures instead of pairing deeper outcomes with the restored shallower ranking, which is the reported contradiction. Memory paths all check out: batch path stable-partitions complete→cut so the cut rows sit at the tail and get freed there while the complete prefix survives; live path frees the deadline-cut candidate's rows on the break; accept upserts the prefix (moving row ownership into `oc_store`) then frees only the array; rollback `destroy_array`s the prefix. And nice that it ships with `test_peg_discarded_stage_outcomes` reproducing the interrupt-at-one-candidate case — that's the hard part to regression-guard. One inline note on where a comment would help; otherwise LGTM.

src/impl/peg.c
olaugh: Added in 4e845503.

The note sits right at the `peg_cand_outcomes_destroy_array` call and states the invariant across all three exits rather than just this one, since that whole-picture view is the part no single hunk shows:

- the completed prefix `[0, done_count)` reaching the rollback was never upserted — accept is the only publisher;
- the cut tail was already freed by whichever evaluation path cut it (live frees the deadline-cut candidate on its `break`, batch frees `[done_count, eval_count)` after stable-partitioning);
- accept upserts exactly `[0, done_count)` and then frees only the array.

So every `rows` allocation is freed exactly once regardless of which exit runs.

Comment only, no behavior change. CI green.

jvc56 (approved):

PR #610, opened 2026-07-22
WordInfoTable

olaugh:
## What this does

Two independent move-generation changes, both move-for-move identical to `main`.

**1. Read board squares directly.** `MoveGen` used to copy each row's `Square`s into a private row cache (`board_copy_row_cache`) before generating; every `gen_cache_get_*` accessor then read the copy. The copy was a win when `Square` was small and no longer is, so the accessors now read the board's own lanes and the per-row `memcpy` is gone. This part needs no data file and is on regardless of `-wit`; it is what the "WIT off" table below measures.

**2. WordInfoTable prunes.** A **WordInfoTable** (WIT) answers one question per board block: *which letters can appear in a word of length L that contains this block?* It is one trie per block length; a block's terminal node holds a row of letter-set bitvectors, one per total word length. Built from a KWG with `convert kwg2wit <lexicon>`, loaded with `-wit true`, off by default.

```c
// Returns the value row for `block` (a word of length `len`), or NULL if
// `block` is not a real word of that length. Entry i is the union of the
// letter sets of all length-(len + i) words containing `block` as a
// contiguous substring.
static inline const uint32_t *word_info_table_lookup(const WordInfoTable *wit,
                                                     const MachineLetter *block,
                                                     int len);
```

**Rows are cached on the board.** The cross-set pass already walks every block, so it stores each block's row on the `Board`, parallel to `squares[]`, keyed at the block's leftmost square in its direction. Move generation reads a lane of those rows with no per-call lookup.

**Prune 1, whole anchors in the WMP generator.** For a playthrough anchor, AND-fold every block's row at this word length into one `addable` set. Every placed tile must be in it, so count the rack tiles that are not and skip the anchor if too few usable tiles remain:

```c
if (gen->word_info_table != NULL && anchor->playthrough_blocks > 0) {
  const uint32_t addable = wgen->playthrough_addable & letter_universe;
  if (addable == 0) {
    return;
  }
  const uint32_t rack_present = (uint32_t)gen->rack_cross_set & letter_universe;
  uint32_t forbidden = rack_present & ~addable;
  int forbidden_count = 0;
  while (forbidden != 0) {
    const MachineLetter forbidden_ml = (MachineLetter)lowest_set_bit_index(forbidden);
    forbidden_count += rack_get_letter(&gen->player_rack, forbidden_ml);
    forbidden = clear_lowest_set_bit(forbidden);
  }
  if (gen->number_of_letters_on_rack - forbidden_count < anchor->tiles_to_play) {
    return;
  }
}
```

**Prune 2, shadow early-stop.** While shadowing rightward into a block, if the tile just placed can never be an addable letter of that block at any reachable word length, no rightward continuation forms a word, so stop. A blank counts as able to realize any addable letter, so no legal play is dropped:

```c
static inline bool wit_shadow_right_block_dead(const MoveGen *gen, int block_col,
                                               uint64_t possible_letters_here,
                                               int min_word_length) {
  const uint32_t *block_row = gen->wit_row_lane[block_col];
  if (block_row == NULL) {
    return false; // uncached or no table: permit all
  }
  const int block_len = gen->wit_len_lane[block_col];
  const bool blank_here = (possible_letters_here & 1U) != 0;
  const uint32_t cand = (uint32_t)possible_letters_here & ~1U;
  uint32_t geq = 0;
  for (int idx = max(min_word_length - block_len, 0); idx < BOARD_DIM - block_len + 1; idx++) {
    geq |= block_row[idx];
    if ((cand & geq) != 0 || (blank_here && geq != 0)) {
      return false;
    }
  }
  return true;
}
```

**Keeping the cache alive across copies.** Rows are derived from the squares, so a reused destination must never keep stale rows, and a rollout copy must not throw the source's rows away either. `board_copy` copies the rows whenever either side has any; when neither does, it skips the cache region and the WIT-off path pays nothing.

**Pairing the table with its KWG.** A `.wit` is found by lexicon name like the KWG and WMP, but it gates correctness. Each table carries an FNV-1a hash of the KWG it was built from, and `set` refuses a table whose hash does not match that player's KWG, naming the file and the rebuild command.

`make release` builds `data/lexica/CSW24.wit` when absent and trains the PGO profile with WIT on. Based on `main`; supersedes #557. #616 builds on this branch.

## Correctness

- Move-for-move sweep with WIT on versus off on the same binary (`witdiff`: 32 seeded CSW24 games, 723 positions): identical move sets, scores and equities, including after board copies and play/undo. The `witsweep` suite runs this comparison in CI on the small fixtures.
- Cache invariants each have a regression test that failed before its fix: a reused copy destination holding a longer block (`8G CAT` reappearing), incremental undo restoring squares without the parallel cache, and `game_update` regenerating rows after a lexicon or WIT change for either player.
- A short shadow anchor inside a longer cached block cannot index past the row (`anchor->word_length - block_len` negative), covered for WIT on and off.
- With WIT off, move generation is handed no row lane, so nothing reads borrowed table storage.
- Dev/ASan suites `wit`, `witcache`, `witsweep`, `wmg`, `shadow`, `movegen`, `sim` pass; cppcheck 2.17.1, circular-dependency check and `format.py` clean.

## Performance

Simulation throughput, same harness on both sides: Apple M4 Mac mini, Apple Clang 17, `-O3 -flto -march=native` (`no_pgo_release`), CSW24, WMP on, 2 plies, 0.5 s/turn, `-minplayiterations 100000`, eight workers via `autoplay games 1 -mtmode igp`; 6 seeds × 2 repetitions × RIT on/off = 24 matched pairs per comparison, order reversed on the repeat; paired geometric mean of iterations/sec, 95 % bootstrap over seed-level means. Host was not fully quiet: `photolibraryd` sat near 25 % of one core throughout, and `fileproviderd` near 100 % of one core during the WIT-off comparison; the paired, order-alternated design absorbs steady load, and the sign tests below are unanimous.

**Whole PR, WIT on, versus `main` (a6c4d6d3):**

| Configuration | Pairs | Effect | 95 % interval |
|---|---|---|---|
| RIT enabled | 12 | +2.01 % | +1.56 % to +2.59 % |
| RIT disabled | 12 | +1.88 % | +1.49 % to +2.32 % |
| Pooled | 24 | **+1.94 %** | +1.54 % to +2.47 % |

Candidate faster in 24/24 pairs (sign test p = 1.2e-07); order effect +0.08 %.

**WIT off versus `main`** (change 1 alone; the `.wit` is not loaded):

| Configuration | Pairs | Effect | 95 % interval |
|---|---|---|---|
| RIT enabled | 12 | +1.49 % | +1.24 % to +1.75 % |
| RIT disabled | 12 | +1.36 % | +1.15 % to +1.58 % |
| Pooled | 24 | **+1.42 %** | +1.26 % to +1.57 % |

Candidate faster in 24/24 pairs (sign test p = 1.2e-07); order effect +0.13 %.

**WIT on versus off, same binary** (what change 2 adds):

| Configuration | Pairs | Effect | 95 % interval |
|---|---|---|---|
| RIT enabled | 12 | +0.51 % | -0.23 % to +1.39 % |
| RIT disabled | 12 | +0.55 % | +0.13 % to +0.98 % |
| Pooled | 24 | **+0.53 %** | -0.03 % to +1.17 % |

WIT on faster in 22/24 pairs (sign test p = 3.6e-05); order effect +0.18 %. A macOS XProtect scan ran during 15 of these 48 runs, which is why the RIT-enabled interval is the widest of the three. WIT itself is worth about half a percent here; most of the PR's gain is change 1.

Profile of the candidate under the same benchmark (RIT on, `sample`, self-time share), WIT off → on: `wmp_get_word_entry` 18.6 % → 17.3 % (the whole-anchor prune skips WMP probes), `game_gen_cross_set` 2.2 % → 3.0 % (filling the block rows), `_platform_memmove` 2.3 % → 2.9 % (`board_copy` now carries the populated rows); the shadow functions are unchanged. Caching per block instead of per side is the obvious next step; #616 builds on this branch and is measured separately.


olaugh:
Correctness / future-proofing note: the WIT block cache (`Board.wit_block_rows` / `wit_block_lens`) lives outside `Square`. `game_gen_cross_set_tracked()` mutates that cache during incremental play, while `MoveUndo` snapshots and restores only `Square`, so `unplay_move_incremental()` does not restore the WIT entries. If a child extends a playthrough block, undo can leave the child's WIT row and block length cached at the parent position.

I audited the current reachability across the endgame variants. Every movegen call *inside the exact endgame solver* currently has WMP disabled:

- `MOVE_RECORD_ALL_SMALL` and `MOVE_RECORD_TILES_PLAYED` disable WMP directly.
- The `MOVE_RECORD_BEST_SMALL` fallback always receives a non-null `solver_get_pruned_kwg()`—either the pruned KWG or the effective full/PEG override KWG—which also disables WMP.
- The single-tile fast path bypasses WMP.

The broader statement that WMP is unused in all endgame-position movegen would be too strong, though. PEG's ordinary root/nested candidate generation and greedy/pessimistic rollouts use WMP-capable `BEST`/`ALL` movegen, including when operating at bag-empty positions. Those paths play on duplicated games rather than using `play_move_incremental()` / `MoveUndo`, so I do not see a current WIT consumer reading the stale cache produced by this gap.

That makes this latent rather than a demonstrated current mis-search, but the `Board`/`MoveUndo` invariant is incomplete and fragile. We should either save and restore the affected WIT row/length entries in `MoveUndo`, or explicitly invalidate/recompute them on undo, and add a play -> unplay cache-equality regression (plus WIT-on versus WIT-off coverage).

olaugh (commented):
Reviewed the WIT-prune-on-PGO speedup. The soundness reasoning on both new prunes is genuinely careful — the comments spell out the blank-as-wild and "never prunes a legal play" arguments, which is exactly what you want for a generation prune. Two notes inline plus one meta:

**Meta / reviewability:** this is stacked on #605 (`pgo-build-and-validation`) and layers on the #557 WIT prunes, so the reviewable delta here is the exact-length anchor prune, the cumulative-length shadow early-stop, the direct board-square reads, and the PGO-trains-with-`-wit` wiring. Reads fine assuming #605 lands first.

**`.wit` staleness:** nice that there's a `WIT_VERSION` guard for on-disk *format* changes. The remaining exposure is *content* staleness — a name-matched `CSW24.wit` built from a different word list than the loaded KWG would silently mis-prune (same name-keyed assumption as KWG/WMP, but now it gates movegen *correctness* rather than just perf). Probably fine given the project-wide convention, but worth being conscious that the Makefile auto-create only triggers when the file is absent, not when it's out of date.

src/impl/move_gen.c
```diff
@@ -915,14 +915,47 @@ void wordmap_gen(MoveGen *gen, const Anchor *anchor) {
...
+  // Word info table prune, lifted entirely out of the loops below. The block
+  // scan in wmp_move_gen_set_playthrough_bit_rack already AND-folded each
+  // playthrough block's letter set at this word length into
+  // wgen->playthrough_addable: every placed tile must be one of those letters.
+  // So the rack must supply tiles_to_play tiles that are each addable (blanks
+  // are wild). Count the rack tiles that are NOT placeable -- present non-blank
+  // letters outside addable, usually just a couple of set bits -- and skip the
+  // whole anchor if too few remain. The blocks (hence addable) are the same for
+  // every start column, so this rules out every candidate play at once.
+  if (gen->word_info_table != NULL && anchor->playthrough_blocks > 0) {
```
olaugh, line 937: The soundness argument for this exact-length anchor prune is convincing (placeable = blanks + addable rack letters; skip when that can't reach `tiles_to_play`). What I couldn't find is a **move-list equivalence** test for the new prunes — `word_info_table_test.c` validates the table itself (lookups vs reference, build/load roundtrip via `assert_wits_equal`), but not that WIT-on generation produces the *same move set* as WIT-off. Since both new prunes are load-bearing-if-wrong and only argued correct in comments, a sweep — random positions × racks, assert `generate_moves` with `-wit true` yields an identical move list to `-wit false` — would turn "never prunes a legal play" from a comment into a guarantee, and would also have caught this class of bug in #557. Does CI already run the movegen suite in both `-wit` modes? If so, worth calling out; if not, that sweep is the highest-value addition here.

src/impl/move_gen.c
```diff
@@ -1982,6 +2060,19 @@ static inline void shadow_play_right(MoveGen *gen, bool is_unique) {
       wmp_move_gen_increment_playthrough_blocks(&gen->wmp_move_gen);
     }
 
+    // WordInfoTable shadow early-stop: if the tile just placed can never be an
```
olaugh, line 2063: The blank handling here is the subtlest part and the comment nails *why* it has to read blank-placeability from `possible_letters_here` rather than the live rack (the blank's already been consumed into the square by this point). One thing worth a second pair of eyes before merge: the conservative treatment assumes a placeable blank can realize *any* addable letter, so the step is only declared dead when the addable set is empty. That's sound for not-dropping-plays, but it also means a square that can only take a blank never gets pruned by this path regardless of the block's addable set — which is the intended trade (correctness over prune-rate), just flagging it's a deliberate softness, not a missed prune.

olaugh:
Added `82f356df` to keep the WIT prune in bounds when a shadow anchor is shorter than the cached board block. In that case `anchor->word_length - block_len` is negative and there is no valid WIT extension row to read; the code now skips only this optional prune and lets normal playthrough validation reject impossible moves. ASAN exposed the out-of-bounds read during the terminal-oracle depth work. A focused regression constructs the short-anchor/long-block case and verifies the WIT mask remains permit-all. Sanitizer-backed `wmg` and `shadow` suites pass after rebasing onto the latest branch tip.

olaugh:
closing this, speedup is gone after fixing the cache bug

olaugh:
Reopening with a fix for why the speedup vanished.

## What was actually wrong

`fca20556` fixed the stale-cache bug by having `board_copy` clear the destination's WIT rows. That is correct, but it also throws away the *source's* rows. `game_copy` runs once per simulation rollout, so every rollout board started with an empty WIT cache and repopulated only the few blocks that incremental play touched. The prunes were correct and almost never had anything to work with — and the copy still paid an 8 KB memset to get there.

Instrumented over one `simbench` game (WIT on, 8 threads):

| | playthrough anchors with a usable forbidden set | subracks skipped |
|---|---:|---:|
| discarding copy (`fca20556`) | 1.41M / 24.5M = **5.7%** | 15.9M / 593M = **2.7%** |
| propagating copy (this push) | 7.00M / 25.5M = **27.4%** | 75.5M / 582M = **13.0%** |

## The fix

The rows are derived from the squares, and the destination receives exactly the source's squares, so the source's rows describe the destination's blocks. `board_copy` now copies the region whenever either side has any row — same memory traffic as the memset it replaces, and the cache survives the copy. When both caches are empty it keeps the existing skip, so the WIT-disabled path pays nothing.

`test_wit_cache_copy_propagates` asserts the destination's rows match the source's byte-for-byte after a copy (and that copying from an empty cache clears them); it fails against the discarding copy. `witdiff` now also plays a move on the copied board and compares against a table-free generator, covering incremental updates layered on an inherited cache — the path the discarding copy made unreachable.

## Benchmarks

Same machine and method as the September 4 numbers above: M4 Mac mini, 8 workers, `no_pgo_release`, `simbench` at 0.5 s/turn, `SIMBENCH_MI=100000`, CSW24 with WMP/RIT, seeds 42/24301/1552, 6 rounds per seed with variant order rotated each round, binaries interleaved. Paired within-round geometric means with a 20,000-resample bootstrap stratified by seed, 18 pairs per row.

| comparison | delta | 95% CI |
|---|---:|---|
| No fix — WIT on vs off (reproduces Sept 4) | −0.54% | [−1.46, +0.40] |
| **With fix — WIT on vs off** (measured with #615 present) | **+0.94%** | [+0.52, +1.52] |
| With fix + #615 + #616 — WIT on vs off | +3.37% | [+2.59, +4.09] |
| **Fix contribution, WIT on** (everything else equal) | **+3.77%** | [+3.41, +4.35] |
| Fix contribution, WIT off (null control) | −0.15% | [−0.61, +0.48] |

Per seed, the fix contribution is +3.91 / +2.90 / +4.49%; no seed disagrees. The September 4 conclusion — WIT on effectively tied with WIT off — was a measurement of the discarding copy, not of WIT.

## Where the remaining cost is

`sample` at 1 ms, profile build, leaf self-time shares, WIT on vs off (both with the fix): `wmp_get_word_entry` 13.55% → 10.98% (−2.56 pp), `wordmap_gen` −1.93 pp, `generate_moves` −0.69 pp; against `game_gen_cross_set` 0.46% → 1.54% (+1.08 pp) and memcpy +0.45 pp. The prune is landing where it was designed to — on the random-access WMP probe. What's left is cache *maintenance*: `game_store_wit_block` does a trie walk per block flanking each regenerated empty square, so a block with empty squares on both sides is looked up twice. That's the next target, as a separate change.

## Validation on this branch

`witcache wmg shadow movegen bitrack` pass; `witcache` passes under ASan/UBSan; `witdiff` (32 seeded CSW24 games, 723 positions, full move/score/equity equality across copy, copy-then-play, and undo) passes; `cppcheck` clean; no circular dependencies; `format.py` clean.

Note on #616: measured separately on the same rounds, its per-subrack prune is +0.27% [+0.08, +0.44] over #615 without this fix and +1.87% [+0.98, +2.55] with it — so its "+3.9%" was measured against the stale cache. It also costs about half a percent with WIT off (−0.52% [−1.22, +0.17]), which it should fix before landing; details there.

src/impl/move_gen.c
olaugh: Added: `witsweep`, registered in the CI `rest` shard (9783e64f, in this push). It builds a WordInfoTable from the loaded CSW21 KWG in memory — no production `.wit` needed — and asserts that WIT-on generation yields the identical complete move list (`MOVE_RECORD_ALL`, `-numplays 10000`) to WIT-off at every position of seeded games, for both players, and again on a copied board after a further play. 4 games / 87 positions: 6.6 s under ASan/UBSan, 0.9 s in release; `WIT_SWEEP_GAMES` lengthens it on demand.

The long form is `witdiff` (on demand, needs `CSW24.wit`): 32 games, 723 positions, same equality, also across copy, copy-then-play and undo.

To the direct question: no, CI did not run the movegen suite in both `-wit` modes before this. It does now, via `witsweep`.

src/impl/move_gen.c
olaugh: Confirmed deliberate — and now covered rather than only argued. `witsweep` and `witdiff` run real racks with blanks against a table-free reference, and #616's `wit_subrack_prune_movegen_test` has explicit blank-plus-forbidden-letter cases (`B?X`, `?DX`, `?DXG`) for the per-subrack prune, which keeps the same rule: a blank is never forbidden.

The tightening you describe is available if wanted: a blank on this square can only realize letters in the square's *cross set*, so `blank_here && (cross_set & geq) != 0` would prune more than `blank_here && geq != 0`. It needs the raw cross set rather than `possible_letters_here` (already rack-masked), and it is a prune-rate change rather than a correctness one, so I've left it as a follow-up instead of folding it into this PR.

olaugh:
Addressing the review notes (b8e25bdf):

**Reviewability.** The description now opens with *What this does* — the table, the board cache, both prunes and the copy semantics, each with code — and *What this enables* with per-step numbers.

**`.wit` content staleness — fixed rather than left to convention.** Each table is now stamped with an FNV-1a hash of the KWG node array it was built from (`kwg_get_hash`, carried in a v3 header), and every `set` validates each player's table against that player's KWG after the WIT — the last data type set — is in place, and after a WIT or KWG reload. A mismatch fails the set, naming the file to delete and the `convert kwg2wit` command to rebuild it. A stale v2 file now fails loudly as well (`detected wit version 2 but only 3 or greater is supported`). Cost: a few milliseconds per `set`. Tables built from a bare word list carry no hash and are not checked.

What it does *not* do: the Makefile still auto-creates only when the file is absent, so an existing out-of-date `data/lexica/CSW24.wit` must be deleted once (`make release` then rebuilds it) — a one-time step for anyone who built this branch before today.

**Tests.** `test_wit_cache_kwg_mismatch` writes one fixture lexicon's table under another lexicon's name and asserts the set is refused with `ERROR_STATUS_WIT_KWG_MISMATCH`; the fixture tables are now built from their KWGs, so the existing lexicon-transition and reload tests exercise the positive check. The `wit` roundtrip test compares the hash.

Both inline threads are answered and resolved.

jvc56 (approved):

PR #611, opened 2026-07-23
Speed up tile-playability move generation

olaugh:
## Summary

- add a score-free `MOVE_RECORD_TILES_PLAYED` traversal that discovers the
  union of playable rack tile types without building strips, scoring moves, or
  inserting moves into a list
- stop as soon as every tile type on the rack has been proven playable
- avoid redundant rack-total mutations in the specialized traversal
- exploit the serialized KWG's alphabetical sibling order: fixed-board-letter
  and general transition scans stop once they have passed the requested letter
- add an isolated benchmark that accepts only verified stuck-tile endgame
  positions and checks the result against the union from
  `MOVE_RECORD_ALL_SMALL` before timing

## Stack

This draft is stacked on #610. It should be reviewed and merged after that PR,
or rebased once #610 lands.

## Why

Endgame stuck-rack detection asks move generation which tile types occur in any
legal move. The old path performed nearly all normal move-generation work and
discarded the resulting moves. Stuck-tile positions are the expensive case:
one or more target tile types are never found, so the query must exhaust the
legal move space.

The specialized traversal keeps only the state required to prove tile
playability. The KWG scan changes also benefit ordinary move generation and are
valid because serialized sibling lists are alphabetically ordered.

## Validation

- exact tile-bitvector comparison against `MOVE_RECORD_ALL_SMALL` for both
  players in 50 verified stuck-tile positions, including seeded
  `initial_tiles_bv` coverage
- KWG, move-generation, endgame, and PEG test suites
- fresh end-to-end `make release` from the final source and profile
- `python3 format.py`
- `git diff --check`

The aggregate test suite's external Google Sheets fetch remains unavailable in
the local environment; the targeted suites above pass.

## Verified benchmarks (post-rebase, board_copy fix included)

Standardized `simbench` steady-state throughput (8-core Apple laptop, 8 threads, RIT on, plies 2, `SIMBENCH_MI=100000`, ~60s/run), `no_pgo_release` builds on every branch (eliminates PGO code-layout variance), binaries interleaved per round, **paired same-round deltas vs the parent branch**, median of 3+ rounds. The whole stack is rebased onto #610 including its `board_copy` WIT-region fix. Chain medians (WIT on): #610 12,085 → #616 13,248 iters/sec (**+9.6% cumulative**).

**This PR** (vs #610): **tied** — simbench paired deltas +11/−50/−58 (median ~0); fixed-work endgame batteries also tie (`benchfp` 500 stuck positions: +0.23%; `benchns` 500 regular: +0.33%). The specialized `MOVE_RECORD_TILES_PLAYED` traversal is algorithmically justified (score-free, threshold early-exit) and correctness-tested, but no end-to-end speedup is claimable at these scales.


src/ent/kwg.h
```diff
@@ -73,10 +73,12 @@ static inline uint32_t kwg_get_next_node_index(const KWG *kwg,
...
   while (1) {
     const uint32_t node = kwg_node(kwg, i);
-    if (kwg_node_tile(node) == letter) {
+    const MachineLetter node_tile = kwg_node_tile(node);
+    if (node_tile == letter) {
       return kwg_node_arc_index_prefetch(node, kwg);
     }
-    if (kwg_node_is_end(node)) {
+    // Serialized sibling lists are alphabetically ordered.
+    if (node_tile > letter || kwg_node_is_end(node)) {
```
olaugh, line 81: This `node_tile > letter` early-out is the load-bearing change: it assumes sibling lists are strictly **ascending** by tile. That holds for `kwg_maker` output, but it's now a hard invariant across *every* KWG the engine loads — wordpruned KWGs, externally converted ones, `override_kwg`, the believed-lexicon KWGs. If any of those ever presents children in a different order (e.g. a frequency-ordered variant), this silently drops letters from generation rather than failing loudly. Worth either a one-line assertion where KWGs are loaded/built that siblings are ascending, or at minimum a comment marking this as a required invariant so nobody reorders children later without realizing movegen correctness now depends on it.

src/impl/move_gen.c
```diff
@@ -1387,6 +1401,153 @@ static inline void go_on_small(MoveGen *gen, int current_col, MachineLetter L,
...
 }
 
+static inline void record_tiles_played(MoveGen *gen, uint64_t played_tiles_bv) {
+  gen->tiles_played_bv |= played_tiles_bv;
+  if ((gen->tiles_played_bv & gen->target_tiles_bv) == gen->target_tiles_bv) {
+    gen->threshold_exceeded = true;
+  }
+}
+
+static inline void recursive_gen_tiles_played(MoveGen *gen, int col,
```
olaugh, line 1411: This is now the ~fourth near-copy of the sibling-scan that got the `node_tile > raw` early-out (`kwg_get_next_node_index`, `recursive_gen`, `recursive_gen_small`, and here). A future change to the scan has to find and update all of them in lockstep. `kwg_get_next_node_index` already encapsulates exactly this "find child == letter, stop once past it" scan — could the fixed-board-letter branch call it instead of re-inlining? If the inlining is load-bearing for perf, a single shared `static inline` helper would at least keep the copies from drifting.

src/impl/move_gen.c
```diff
@@ -3217,6 +3378,41 @@ void gen_record_scoring_plays_small(MoveGen *gen) {
   }
 }
 
+static void gen_record_tiles_played(MoveGen *gen) {
```
olaugh, line 3381: This specialized path replaces the old `gen_record_scoring_plays_small` route for `MOVE_RECORD_TILES_PLAYED`, but the only equivalence test (`board_test.c:545` `test_board_get_playable_tiles_bv`) exercises **single-tile** racks — which never reach the multi-tile recursion, the `threshold_exceeded` early-out, or the blank+letter interplay that make up most of this new code (and it's exactly where the sorted-sibling early-out would bite if the invariant were ever violated). Recommend a multi-tile equivalence test: for a handful of random positions × multi-tile racks, compare this bitvector against the tile-type union derived from a full `MOVE_RECORD_ALL_SMALL` generation. That pins both the new traversal and the early-out in one shot.

olaugh (commented):
Reviewed the tile-playability movegen speedup — clean, well-contained optimization and the sorted-sibling early-out is a nice win. I also pushed a commit fixing the clang-tidy CI failure (`benchmark_endgame_test.c` was using `MachineLetter` with only a transitive include). Three non-blocking things worth a look inline: (1) the sorted-sibling invariant the early-out now silently depends on, (2) the growing duplication of the sibling-scan loop, and (3) multi-tile correctness coverage for the new specialized path.

PR #612, opened 2026-07-23
Avoid materializing losing WMP moves

olaugh:
## What this does

In the WMP record path, every candidate word was written out as a full `Move` — strip, position, tiles — before its equity was computed and compared against the cutoff, even when it was about to be discarded. On a nonempty board the parts of static equity that matter for that comparison do not depend on placement: score, leave value, the number of tiles played, the bag, and the racks. (The opening move's placement penalty does, so an empty board keeps the old path.)

This PR computes that equity first, from values the generator already has, and only materializes the `Move` when it can still matter:

- `MOVE_RECORD_BEST`: skip a candidate whose equity is below the best so far.
- `MOVE_RECORD_ALL` with a full list: skip a candidate below the list's minimum.
- `MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST`: skip a candidate below the cutoff.

Every skip is on a strict `<`, so ties still go through the full comparison and tie-breaking is unchanged. The shared arithmetic moves into `static_eval_get_nonopening_move_equity`, which the existing `Move`-based path now calls too, so the two cannot drift. The tile count it uses is `gen->max_tiles_to_play`, which is also what `set_play_for_record_wmp` writes into the `Move`; the leave and opponent rack are the same objects the `Move` path reads. In `MOVE_RECORD_BEST` the inference threshold is only checked when the best move is replaced, which a skipped candidate never does; the other two modes leave inference (`stop_on_threshold`) on the original path.

Based on `main`, which now includes #614 (merged as `fb0fc7f7`), so the pre-check sits inside the record loop #614 specializes per generation.

## Performance

Against `main` with #614 merged (measured at `6847ade5`, #614's final commit, which is what `main` merged), WIT on: M4 Mac mini, 8 workers, `no_pgo_release`, `simbench` 0.5 s/turn, seeds 42/24301/1552 × 6 interleaved rounds, paired within-round geometric means, stratified-bootstrap 95% CIs, 18 pairs per row. All three seeds are positive in both rows.

|---|---:|---:|
| **RIT on** (the production configuration) | **+0.80%** | [+0.68%, +0.91%] |
| RIT off | +0.63% | [+0.51%, +0.74%] |

**Why it pays.** Counting in the same workload (one 20 s `simbench` run, 8 threads): in the playouts (`MOVE_RECORD_BEST`) the shortcut discards **90.8%** of WMP candidate words before a `Move` is written — 25.1M of 27.6M with the RIT on, 90.7% with it off — and 95.5% at the top level (capacity-bounded `MOVE_RECORD_ALL`). In a `BUILD=profile` sample (inlining suppressed, RIT on) the record-path leaves — `record_wmp_play`, `record_wmp_plays_for_word`, `static_eval_get_move_equity_with_leave_value` — go from 0.59% of samples to 0.39%, and `get_wmp_equity_without_move` costs 0.05%; the shortcut removes roughly a third of what the record path spent, and nothing else in the profile moves.

## Correctness

- **New test** `test_wmp_bounded_record_modes` (`wmg` suite, runs in CI): on 11 CSW21 positions — midgame boards, racks with blanks and fewer than seven tiles, a short bag, an empty bag, and an empty board where the shortcut declines — in both sort types. It first pins the WMP generator's full list to the generator without a WMP, whose equities come from the materialized `Move`, so the precomputed value is checked against the real one and not only against itself. It then checks that `MOVE_RECORD_BEST` returns the equity of the head of the full list, that capacity-1/3/10 lists carry the equities of its first *k* entries, and (equity sort) that within-0/5/20/40 lists hold every entry strictly inside the margin, every bounded move being one the full list holds at that equity. It passes on `main` without this PR too. Mutation check: an equity one point low, a tile count one low, and each skip made one point too eager (`BEST`, full `ALL`, within-x) are all caught; relaxing a skip's strict `<` to `<=` is not, because that only changes tie-breaking, which the test tolerates on purpose and the differential below covers.
- **Score sort + within-x + WMP**: while writing the test I found that a score-sorted `MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST` list from the WMP generator drops some vertical plays, including ties for the best (e.g. WINDY board, rack ADEEGIL, margin 20: E7 LID for 4 is missing; margin 5 on another position loses a 24-point tie for the top). The generator without a WMP is complete, and it happens identically on `main`, so it predates this PR; the test limits the within-x check to equity sort. Tracked in #673 and fixed in #674; once that merges, the guard can go.
- **Autoplay differential** against `main`: the same seeded 200 game pairs (400 games) played by `main`'s `no_pgo_release` binary (with #614 merged) and this one, CSW24 with WMP and WIT, in four record/sort configurations (`best`/equity, `all -numplays 15`/equity, `best`/score, `all -numplays 5`/score), each with the RIT on and off: all eight aggregate reports (turns, scores, bingos, end reasons) are identical.
- Earlier, before #614 landed, a move-for-move differential against the `main` of that time (32 seeded CSW24 games, 712 positions, 967,104 moves per export, RIT on and off) and a Wordsmog differential (20,027,705 moves) were identical.
- Release suites `wmg movegen shadow rit witsweep` and the WIT copy differential on this tree; ASan `wmg movegen`; cppcheck, circular-dependency check and `format.py` clean.

## What can build on this

- **Fold the switches to constants.** `get_wmp_equity_without_move` switches on `move_sort_type` per candidate, and the record path on `move_record_type`. Both are fixed for a generation. #614's conditional-inlining pattern — an `always_inline` body with a `bool`/enum every caller passes as a constant, instantiated once per mode by a per-generation dispatch — would make both branches vanish from the hot path. Untested; the pattern's cost model is only known for the RIT switch.
- **#614** (merged into `main`): lazy resolution of RIT-backed WMP entries — the other half of the record path. The number above is this PR's increment on top of it.
- **#618–#623** (pending drafts): WMP shadow-side work stacked on the merged `wit-subrack-pruning`, each needing a `rebase --onto main`; #618 measured at the noise floor.

### DIFF
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -669,6 +669,36 @@ static inline Equity get_move_equity_for_sort_type_wmp(MoveGen *gen,
 #endif
 }
 
+// On a nonempty board, static equity depends only on the score, leave, racks,
+// and number of tiles played. Compute it before writing a full Move so bounded
+// record modes can discard a losing candidate without copying its strip.
+// Opening placement penalties inspect the Move and use the regular path
+// instead.
+static inline bool get_wmp_equity_without_move(const MoveGen *gen, Equity score,
+                                               Equity leave_value,
+                                               Equity *equity) {
+  switch (gen->move_sort_type) {
+  case MOVE_SORT_SCORE:
+    *equity = score;
+    return true;
+  case MOVE_SORT_EQUITY:
+    if (gen->board_number_of_tiles_played == 0) {
+      return false;
+    }
+    *equity = static_eval_get_nonopening_move_equity(
+        &gen->ld, &gen->leave, &gen->opponent_rack, gen->max_tiles_to_play,
+        gen->number_of_tiles_in_bag, score, leave_value);
+    return true;
+  default:
+    log_fatal("unhandled move sort type: %d", gen->move_sort_type);
+  }
+#if defined(__has_builtin) && __has_builtin(__builtin_unreachable)
+  __builtin_unreachable();
+#else
+  return false;
+#endif
+}
+
 static inline void
 update_best_move_or_insert_into_movelist_wmp(MoveGen *gen, int start_col,
                                              Equity score, Equity leave_value) {
@@ -677,10 +707,28 @@ update_best_move_or_insert_into_movelist_wmp(MoveGen *gen, int start_col,
   switch (gen->move_record_type) {
   case MOVE_RECORD_ALL:
   case MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST: {
+    Equity precomputed_equity = 0;
+    const bool has_precomputed_equity = get_wmp_equity_without_move(
+        gen, score, leave_value, &precomputed_equity);
+    if (has_precomputed_equity && !gen->stop_on_threshold) {
+      if (gen->move_record_type == MOVE_RECORD_ALL &&
+          move_list_get_count(gen->move_list) ==
+              move_list_get_capacity(gen->move_list) &&
+          precomputed_equity < move_list_peek_equity(gen->move_list)) {
+        return;
+      }
+      if (gen->move_record_type == MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST &&
+          gen->best_move_equity_or_score != EQUITY_INITIAL_VALUE &&
+          precomputed_equity < gen_get_cutoff_equity_or_score(gen)) {
+        return;
+      }
+    }
     Move *move = move_list_get_spare_move(gen->move_list);
     set_play_for_record_wmp(gen, move, start_col, score);
     move_equity_or_score =
-        get_move_equity_for_sort_type_wmp(gen, move, leave_value);
+        has_precomputed_equity
+            ? precomputed_equity
+            : get_move_equity_for_sort_type_wmp(gen, move, leave_value);
     if (gen->move_record_type == MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST) {
       // This updates the cutoff move internally so no update will be pending
       // afterward.
@@ -691,10 +739,19 @@ update_best_move_or_insert_into_movelist_wmp(MoveGen *gen, int start_col,
     break;
   }
   case MOVE_RECORD_BEST: {
+    Equity precomputed_equity = 0;
+    const bool has_precomputed_equity = get_wmp_equity_without_move(
+        gen, score, leave_value, &precomputed_equity);
+    if (has_precomputed_equity &&
+        precomputed_equity < move_get_equity(gen_get_readonly_best_move(gen))) {
+      return;
+    }
     Move *current_move = gen_get_current_move(gen);
     set_play_for_record_wmp(gen, current_move, start_col, score);
     move_equity_or_score =
-        get_move_equity_for_sort_type_wmp(gen, current_move, leave_value);
+        has_precomputed_equity
+            ? precomputed_equity
+            : get_move_equity_for_sort_type_wmp(gen, current_move, leave_value);
     move_set_equity(current_move, move_equity_or_score);
     if (compare_moves(current_move, gen_get_readonly_best_move(gen), false)) {
       need_to_update_best_move_equity_or_score = true;
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -672,6 +672,36 @@ static inline Equity get_move_equity_for_sort_type_wmp(MoveGen *gen,
 #endif
 }
 
+// On a nonempty board, static equity depends only on the score, leave, racks,
+// and number of tiles played. Compute it before writing a full Move so bounded
+// record modes can discard a losing candidate without copying its strip.
+// Opening placement penalties inspect the Move and use the regular path
+// instead.
+static inline bool get_wmp_equity_without_move(const MoveGen *gen, Equity score,
+                                               Equity leave_value,
+                                               Equity *equity) {
+  switch (gen->move_sort_type) {
+  case MOVE_SORT_SCORE:
+    *equity = score;
+    return true;
+  case MOVE_SORT_EQUITY:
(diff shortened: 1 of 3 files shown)


src/impl/move_gen.c
```diff
@@ -669,6 +669,36 @@ static inline Equity get_move_equity_for_sort_type_wmp(MoveGen *gen,
...
+  switch (gen->move_sort_type) {
+  case MOVE_SORT_SCORE:
+    *equity = score;
+    return true;
+  case MOVE_SORT_EQUITY:
+    if (gen->board_number_of_tiles_played == 0) {
+      return false;
+    }
+    *equity = static_eval_get_nonopening_move_equity(
+        &gen->ld, &gen->leave, &gen->opponent_rack, gen->max_tiles_to_play,
```
olaugh, line 689: The soundness of the whole optimization rests on `gen->max_tiles_to_play` equaling each candidate's actual `move_get_tiles_played(move)` — which is true today because the WMP anchor sets it to `anchor->tiles_to_play` (line 938) and all of an anchor's plays use that many rack tiles. It only matters in the small-bag regime (where `bag_plus_rack_size < PEG_ADJUST_VALUES_LENGTH`), so a future change that let a WMP anchor emit variable-length plays would corrupt near-endgame equities *silently* and only there. Worth a one-line comment either here or at line 938 pinning that invariant, since it's load-bearing and invisible in the common midgame path.

olaugh (commented):
Reviewed the "avoid materializing losing WMP moves" speedup — this one's clean. The key design choice is right: factoring `static_eval_get_nonopening_move_equity` out so the pre-materialization estimate and the full-Move path call the *same* function means the precomputed equity equals the materialized equity by construction, so a candidate can never be wrongly rejected due to a formula drift. I checked the one thing that could still bite — the precompute passes `gen->max_tiles_to_play` where the full path uses `move_get_tiles_played(move)` — and it holds: the WMP anchor sets `max_tiles_to_play = anchor->tiles_to_play` (move_gen.c:938), and every candidate at that anchor plays exactly that many rack tiles, so the peg/endgame adjustment index matches. Tie preservation also looks correct (all three rejections use strict `<`, so equal-equity candidates still materialize for the deterministic tie-break). Also pushed a commit fixing the clang-tidy failure (same transitive-`MachineLetter` include as #611, reintroduced by this stack's benchmark rewrite). Two small notes inline; nothing blocking. Stacked on #611.

src/impl/move_gen.c
```diff
@@ -677,10 +707,28 @@ update_best_move_or_insert_into_movelist_wmp(MoveGen *gen, int start_col,
...
   case MOVE_RECORD_ALL:
   case MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST: {
+    Equity precomputed_equity = 0;
+    const bool has_precomputed_equity = get_wmp_equity_without_move(
+        gen, score, leave_value, &precomputed_equity);
+    if (has_precomputed_equity && !gen->stop_on_threshold) {
+      if (gen->move_record_type == MOVE_RECORD_ALL &&
+          move_list_get_count(gen->move_list) ==
+              move_list_get_capacity(gen->move_list) &&
+          precomputed_equity < move_list_peek_equity(gen->move_list)) {
```
olaugh, line 717: Because the precompute reuses the exact same equity function, this is behavior-preserving by construction rather than an approximation — nice. The one thing I couldn't find is a test that *locks that in*: a WMP-heavy position generated with `MOVE_RECORD_ALL` should yield a byte-identical move list whether or not these early rejections fire. Given the strict-`<` tie handling is subtle (miss it and you'd drop a tie that the full path keeps), a small equivalence test — same position, assert the resulting move list is identical to a build/flag with the rejections disabled — would be cheap insurance and would catch any future refactor that lets the two equity paths drift.

olaugh:
Rebased onto main past #616 — clean, still 2 files. Verified on the rebase: `wmg movegen shadow witsweep` and `witdiff` in release, `wmg movegen` under ASan; cppcheck, circular-deps and `format.py` clean.

Review note: the precomputed equity takes its tile count from `gen->max_tiles_to_play`, which `wordmap_gen` sets to `anchor->tiles_to_play` (`move_gen.c:1038`) — every word for a size-*k* subrack plays exactly *k* tiles, so it matches the count the materialized `Move` would report. Ties still reach the full comparison (the pre-check skips only on strictly-less), and the opening-board case falls back to the regular path.

src/impl/move_gen.c
olaugh: Added in 2d0c7290, at both ends: the comment on `get_wmp_equity_without_move` says the tile count is `gen->max_tiles_to_play`, that `wordmap_gen` sets it to the anchor's `tiles_to_play` (every word for a size-*k* subrack plays *k* tiles, so it is what `set_play_for_record_wmp` writes into the Move), and that the pre-endgame adjustment is the one place a variable-length anchor would break it, silently. The assignment in `wordmap_gen` carries the matching note pointing back here.

src/impl/move_gen.c
olaugh: Locked in by `test_wmp_bounded_record_modes` (`wmg` suite, 06b8f8d5): on 11 CSW21 positions in both sort types it first pins the WMP generator's full `MOVE_RECORD_ALL` list to the generator without a WMP, whose equity comes from the materialized Move, so the precomputed value is checked against the real one and not only against itself; then it checks `MOVE_RECORD_BEST`, capacity-1/3/10 lists, and within-0/5/20/40 lists against that full list. Mutation check: an equity one point low, a tile count one low, and each rejection made one point too eager are all caught.

On the strict-`<` tie handling: the test compares equal-equity moves by membership rather than position, because shadow already decides ties by anchor order across record modes, so relaxing `<` to `<=` is deliberately *not* caught there. That case is covered by the move-for-move differential instead: 200 game pairs in four record/sort configurations, RIT on and off, played by #614's binary and this one, all eight reports identical (a dropped tie changes tie-breaking and so the games).

jvc56 (approved):

PR #614, opened 2026-07-23
Defer RIT-backed WMP lookups until record time

olaugh:
## What this does

With a RackInfoTable loaded, the generator already knows, for the rack on turn, whether a word of each length exists and the best leave for each leave size. It nevertheless looked up the WMP entry of **every** canonical subrack up front, before shadow had run, purely to have the pointer ready in case record time wanted that subrack's words — and shadow prunes most candidates before record time ever asks.

**Lazy resolution.** When the RIT is loaded, a subrack's WMP entry is looked up the first time record time actually needs its words, and the result — entry or miss — is kept in the per-rack cache for the next occurrence of the rack. An entry never needed holds a sentinel pointer (`WMP_ENTRY_UNRESOLVED`), only ever compared, never dereferenced. Under simulation the eager walk was 35% of all WMP hash lookups and deferral removes 88% of those; over an export that records every move it saves almost nothing, because record time then reaches nearly every subrack — the saving is entirely a product of pruning.

**Conditional inlining.** Without a RIT, nothing needs resolving, but a sentinel test in the record loop would still cost. So `lazy` is a compile-time constant: `gen_record_scoring_plays` decides once per generation, on `gen->rit_entry`, between two instantiations of the same `always_inline` anchor loop — an eager copy where the test folds away, and the lazy copy for RIT-backed racks. Neither path pays a per-anchor call or a per-subrack test; one binary serves both.

```c
static inline __attribute__((always_inline)) void gen_record_scoring_plays(MoveGen *gen) {
  if (gen->rit_entry == NULL) {
    gen_record_scoring_plays_eager(gen);       // noinline: _impl(gen, /*lazy=*/false)
  } else {
    gen_record_scoring_plays_impl(gen, true);  // inlined
  }
}
```

The eager path asserts it never meets a sentinel, which holds because the per-thread subrack cache is invalidated whenever the RIT changes (`move_gen.c:3014`).

## Performance

Against today's `main` (post #610/#616, WIT on): M4 Mac mini, 8 workers, `no_pgo_release`, `simbench` 0.5 s/turn, seeds 42/24301/1552 × 6 interleaved rounds, paired within-round geometric means, stratified-bootstrap 95% CIs, 18 pairs per row.

| configuration | this PR vs main |
|---|---:|
| **RIT on** (the production configuration) | **≈ +4.0%** — lazy resolution +3.53% [+3.42, +3.62], and the per-generation specialization a further +0.51% [+0.30, +0.74] over that |
| RIT off | **+0.69%** [+0.46, +0.93] |

Both specialization numbers were first measured in a separate session (+0.41% and +0.54%) and replicated here with an independent harness; every seed positive in both. Lazy resolution alone had cost −0.24% with the RIT off; that is gone. The gain is larger than the ~+1.1% first measured on a pre-WIT `main` because the WIT prunes leave even fewer subracks for record time to resolve.

## Correctness

- Move-for-move differential against `main`, RIT on and off: 32 seeded CSW24 games, 712 positions, 967,104 moves per export, all four sorted exports identical.
- `test_rit_movegen_equality` (`ap_rit` shard, TWL98): a RIT game and a plain game in lockstep, complete sorted move lists compared at every position — 134 positions, ~130k moves. With lazy resolution deliberately broken the RIT-on export collapses to a third of its moves and this test fails on the first position; it asserts the RIT actually loaded, so it cannot pass vacuously.
- `test_rit_toggle_subrack_cache` (`ap_rit`): the same rack generated across `set -rit false` and back in one config, compared against a plain generator, with the RIT-backed generation stopped at its first play so that unresolved entries are what would carry over — pins the cache invalidation the eager path's assert relies on.
- Both under ASan/UBSan on the final layout; `witdiff` (32 games, 723 positions) identical; cppcheck, circular-dependency check and `format.py` clean.

## What can build on this

**The conditional-inlining pattern is reusable.** An `always_inline` body taking a `bool` that every call site passes as a constant, instantiated once per mode by a single dispatch at the top of a generation, gives mode-specialized code with no runtime test in the hot loop and no code duplication in the source. Other per-generation switches that are currently tested per anchor or per move and could be specialized the same way, none yet measured:

- **WIT on/off** — `gen->word_info_table != NULL` is tested in `wordmap_gen` and in the shadow early-stop for every anchor.
- **Sort and record type** — `get_wmp_equity_without_move` and the record path switch on `move_sort_type`/`move_record_type` per candidate move; #612's pre-check would fold to a constant.
- **Wordsmog and the small-move record types** — already dispatched per anchor by branches on `gen->is_wordsmog` and `move_record_type`.

Pending PRs on the same path:

- **#612** — skip materializing a WMP move whose equity already loses; +0.45% replicated, no measured cost. Independent of this PR and the most direct candidate for the specialization above.
- **#618–#623** — WMP shadow-side drafts stacked on the now-merged `wit-subrack-pruning`; each needs a `rebase --onto main` as this PR got. #618 itself measured at the noise floor.

Measured and not worth pursuing: memoizing `game_store_wit_block` would save only ~22% of its lookups (~0.1 pp).

### DIFF
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -3649,12 +3649,8 @@ void generate_moves(const MoveGenArgs *args) {
           leaves_are_populated && subrack_entry->valid &&
           bit_rack_equals(&subrack_entry->key, &wgen->player_bit_rack);
       if (subrack_cache_hit) {
-        // Restore enumerate_nonplaythrough_subracks output AND the
-        // per-subrack wmp_entry pointers from cache. Both pieces are
-        // rack-determined (subracks via combinatoric walk, wmp_entries
-        // via WMP hash), so on hit we skip both the enumeration and the
-        // per-subrack wmp_get_word_entry calls that the size walk would
-        // otherwise run.
+        // Restore the enumeration output and whichever per-subrack WMP
+        // entries prior occurrences of this rack actually needed.
         memcpy(wgen->count_by_size, subrack_entry->count_by_size,
                sizeof(wgen->count_by_size));
         for (int i = 0; i < MOVEGEN_SUBRACK_CACHE_ENTRIES; i++) {
@@ -3663,18 +3659,22 @@ void generate_moves(const MoveGenArgs *args) {
               subrack_entry->leave_values[i];
           wgen->nonplaythrough_infos[i].wmp_entry =
               subrack_entry->wmp_entries[i];
+          wgen->nonplaythrough_infos[i].wmp_entry_is_set =
+              subrack_entry->wmp_entries_are_set[i];
         }
       }
+      if (leaves_are_populated) {
+        wgen->nonplaythrough_wmp_entry_cache = subrack_entry->wmp_entries;
+        wgen->nonplaythrough_wmp_entry_cache_set =
+            subrack_entry->wmp_entries_are_set;
+      }
       if (gen->rit_entry != NULL) {
-        // RIT-backed fast path: skip the per-size wmp_get_word_entry loop
-        // for any played size where the RIT entry says no canonical
-        // k-subrack of this rack forms a k-letter word on its own. Seeds
-        // nonplaythrough_best_leave_values directly from the cached max
-        // the RIT already computed at build time.
+        // The RIT supplies existence and best-leave bounds. Defer each
+        // per-subrack WMP lookup until wordmap generation survives those
+        // bounds and actually needs that subrack's words.
         wmp_move_gen_check_nonplaythrough_existence_with_rit(
             wgen, check_leaves, &gen->leave_map, gen->rit_entry,
-            /*subracks_precomputed=*/subrack_cache_hit,
-            /*wmp_entries_precomputed=*/subrack_cache_hit);
+            /*subracks_precomputed=*/subrack_cache_hit);
       } else {
         wmp_move_gen_check_nonplaythrough_existence(
             wgen, check_leaves, &gen->leave_map,
@@ -3682,8 +3682,8 @@ void generate_moves(const MoveGenArgs *args) {
             /*wmp_entries_precomputed=*/subrack_cache_hit);
       }
       if (!subrack_cache_hit && leaves_are_populated) {
-        // Store the newly-computed enumeration and wmp_entry pointers
-        // into the cache. Only cache when leaves_are_populated so we
+        // Store the newly computed enumeration and any WMP entries that have
+        // already been resolved. Only cache when leaves_are_populated so we
         // don't stash garbage leave_values from an uninitialized leave_map.
         subrack_entry->key = wgen->player_bit_rack;
         subrack_entry->valid = true;
@@ -3695,6 +3695,8 @@ void generate_moves(const MoveGenArgs *args) {
               wgen->nonplaythrough_infos[i].leave_value;
           subrack_entry->wmp_entries[i] =
               wgen->nonplaythrough_infos[i].wmp_entry;
+          subrack_entry->wmp_entries_are_set[i] =
+              wgen->nonplaythrough_infos[i].wmp_entry_is_set;
         }
       }
     }
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -41,6 +41,9 @@
 #include <stdlib.h>
 #include <string.h>
 
+// See WMP_ENTRY_UNRESOLVED in wmp_move_gen.h. Only its address is used.
+const WMPEntry wmp_entry_unresolved_sentinel = {0};
+
 #define INITIAL_LAST_ANCHOR_COL (BOARD_DIM)
 
 // Cache move generators since destroying
@@ -3480,14 +3483,10 @@ void generate_moves(const MoveGenArgs *args) {
               subrack_entry->leave_values[i];
           wgen->nonplaythrough_infos[i].wmp_entry =
               subrack_entry->wmp_entries[i];
-          wgen->nonplaythrough_infos[i].wmp_entry_is_set =
-              subrack_entry->wmp_entries_are_set[i];
         }
       }
       if (leaves_are_populated) {
         wgen->nonplaythrough_wmp_entry_cache = subrack_entry->wmp_entries;
-        wgen->nonplaythrough_wmp_entry_cache_set =
-            subrack_entry->wmp_entries_are_set;
       }
       if (gen->rit_entry != NULL) {
         // The RIT supplies existence and best-leave bounds. Defer each
@@ -3516,8 +3515,6 @@ void generate_moves(const MoveGenArgs *args) {
               wgen->nonplaythrough_infos[i].leave_value;
           subrack_entry->wmp_entries[i] =
               wgen->nonplaythrough_infos[i].wmp_entry;
-          subrack_entry->wmp_entries_are_set[i] =
-              wgen->nonplaythrough_infos[i].wmp_entry_is_set;
         }
       }
     }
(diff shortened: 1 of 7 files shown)


olaugh:
Rebased onto main past #616: one additive struct conflict (#610's `playthrough_addable` and this PR's cache pointers wanted the same spot — both kept) plus test-include ordering; still 3 commits, 7 files. Verified on the rebase: `wmg movegen shadow rit witsweep` and `witdiff` in release, `wmg movegen rit` under ASan; cppcheck, circular-deps and `format.py` clean.

Review: the sentinel is only ever compared, never dereferenced — enumeration marks fresh subracks unresolved (`wmp_move_gen.h:471`), the getter resolves on demand (`:686`), and the eager non-RIT existence walk asserts it never sees one (`:372`).

One caveat worth a number: since #616, `wmp_move_gen_get_subrack_words` is `always_inline` (a second caller had un-inlined it and cost ~5%), and this PR makes that body larger. The −0.3% RIT-off control above predates that, so I'm re-running the paired benchmark on this base (RIT on/off, WIT on) and will post it here.

olaugh:
Paired benchmark on the rebased branch (`158682c9`) against today's main (`94961e9a`, post-#610/#616), **WIT on** for both since the release trains with it: M4 Mac mini, 8 workers, `no_pgo_release`, `simbench` 0.5 s/turn, `SIMBENCH_MI=100000`, seeds 42/24301/1552 × 6 interleaved rounds, order rotated per round, paired within-round geometric means, 20k-resample bootstrap stratified by seed, 18 pairs per row.

| comparison | delta | 95% CI |
|---|---:|---|
| **#614 vs main, RIT on** | **+3.71%** | [+3.57, +3.85] |
| #614 vs main, RIT off (control) | −0.24% | [−0.44, −0.07] |
| main: RIT on vs off | +16.37% | [+16.19, +16.54] |
| #614: RIT on vs off | +20.98% | [+20.72, +21.31] |

Per seed, RIT on: +2.40 / +2.27 / +6.52%. That is a good deal more than the +1.1% measured before #610 and #616 landed, and for the reason this PR's own analysis gives — the saving is a product of pruning, and the WIT prunes now let far fewer subracks reach record time. The RIT-off control reproduces the earlier −0.3%: with the getter `always_inline` since #616, that is one predictable sentinel compare per nonplaythrough subrack in the hot loop. It could be moved out of the table-free path the way #616's mask test was, as a follow-up; at −0.24% against +3.7% on the production configuration it does not block this.

olaugh:
Pushed `9e8a0bdb`: the RIT-off cost is gone, at a measured price on the RIT-on side that is yours to accept or not.

**What changed.** `lazy` is now a parameter of `wmp_move_gen_get_subrack_words` and of `wordmap_gen`'s record helper. Both are `always_inline` and every call site passes a constant, so the in-line loop for the plain case — no WIT forbidden set, no RIT entry — compiles without the sentinel test, exactly as before this PR. RIT-backed racks and masked anchors take the out-of-line scan, which holds the lazy loops. The eager path asserts it never meets a sentinel, which the RIT-change invalidation of the subrack cache (`move_gen.c:3014`) guarantees. A new `test_rit_toggle_subrack_cache` (in `ap_rit`, next to `test_rit_movegen_equality`) pins that invalidation: it generates the same rack across `set -rit false` and back in one config — same WMP object, the cache's other key — stopping the RIT-backed generation at its first play so unresolved entries are what would carry over, and compares complete move lists against a plain generator.

**Numbers** (M4 mini, 8 workers, `no_pgo_release`, WIT on, paired simbench, 18 pairs per row, stratified bootstrap 95% CIs; method as in the #610 reopen comment):

| vs main | before (`158682c9`) | now (`9e8a0bdb`) | two in-line loops (rejected) |
|---|---:|---:|---:|
| RIT off | −0.24% [−0.44, −0.07] | **+0.15% [+0.05, +0.26]** | −0.75% [−0.81, −0.69] |
| RIT on | **+3.71% [+3.57, +3.85]** | +3.16% [+3.02, +3.27] | +3.11% [+3.01, +3.21] |

The 0.55 points given back with the RIT on are the per-anchor call for lazy anchors, which the original paid only for masked anchors. Keeping both unmasked loops in line — the obvious way to avoid it — recovers none of it and costs the eager loop 0.75% (measured, last column), so the dispatch stays out of line. If you would rather keep the full +3.71% on the production configuration and accept −0.24% with the RIT off, the previous head does exactly that; say so and I'll restore it.

Verified on the push: `ap_rit rit wmg` (dev/ASan), `movegen wmg shadow rit witsweep` and `witdiff` (release); cppcheck, circular-deps, `format.py` clean; only the out-of-line scan has a standalone symbol.

olaugh:
Restored the original record loop as `d65cf5a2` (`158682c9` plus the toggle test) — RIT on is the flagship, and a direct check confirms the trade is real but small.

Within one campaign, all three RIT on, WIT on, 18 pairs each:

| vs main | delta | 95% CI |
|---|---:|---|
| original loop (`158682c9`) | **+3.53%** | [+3.42, +3.62] |
| three-loop dispatch (`9e8a0bdb`) | +3.16% | [+3.05, +3.24] |
| original vs three-loop | +0.36% | [+0.23, +0.48] |

Per seed the gap is −0.02 / +0.35 / +0.75%. Against that, the three-loop form's RIT-off advantage was ~0.4% (+0.15% vs −0.24%). Symmetric enough that the production configuration decides it: the original stays, and RIT-off keeps its −0.24% [−0.44, −0.07]. The structures that could have had both were measured and don't — both loops in line costs the eager loop 0.75% and recovers nothing.

So the branch is main + four commits: the three original ones and `test_rit_toggle_subrack_cache`, which pins the subrack-cache invalidation across a RIT switch (`move_gen.c:3014`) that the eager walk's assert relies on. My earlier note about a cache hazard was wrong — that invalidation already covers it.

olaugh:
Following up on the trade-off discussed in [the `9e8a0bdb` comment](https://github.com/jvc56/MAGPIE/pull/614#issuecomment-5562460177) and [the restore](https://github.com/jvc56/MAGPIE/pull/614#issuecomment-5562583057): **no trade-off is necessary.** `6847ade5` gets both.

The choice was never between a sentinel test in the eager loop and a call per anchor. Dispatching once per *generation* — `gen_record_scoring_plays` picks, on `gen->rit_entry`, between two instantiations of the same `always_inline` anchor loop, with `lazy` a compile-time constant in each — removes the test from the eager path and the call from the lazy one. One binary; each mode runs its own specialized loop.

Measured on the same M4, paired within-round, 18 pairs per row, replicating a separate session's result (+0.41% / +0.54%):

| comparison | delta | 95% CI |
|---|---:|---|
| `6847ade5` vs the restored head `d65cf5a2`, RIT on | **+0.51%** | [+0.30, +0.74] |
| `6847ade5` vs main, RIT off | **+0.69%** | [+0.46, +0.93] |

Every seed positive in both. Against main that puts the PR at roughly +4.0% with the RIT on (+3.53% [+3.42, +3.62] from lazy resolution, compounded with the +0.51%) and +0.69% with it off — the −0.24% in [the earlier benchmark](https://github.com/jvc56/MAGPIE/pull/614#issuecomment-5561649925) is gone, and the RIT-on side gained rather than paid. The 0.36% gap the restore accepted was an artifact of dispatching per anchor, which was the wrong level.

Correctness on the final layout: `ap_rit`'s equality and toggle tests under ASan, the 32-game differential, `witdiff`, and the lints; the description now carries the full picture, including where else the same conditional-inlining pattern could apply.

jvc56 (approved):

PR #620, opened 2026-07-23
Avoid sorting and copying shadow multiplier backups

olaugh:
## What this does

Shadow scoring pairs descending tile scores with descending effective letter multipliers to bound the best score for an anchor. It also retains each square's cross-word multiplier and column so it can rebuild that effective list when the main-word multiplier changes.

**Append backup entries.** Only the effective list needs sorting. `insert_unrestricted_multipliers` now appends the backup pair, removing a second sorted insertion from each unrestricted square. `maybe_recalculate_effective_multipliers` still processes every pair and sorts the resulting effective multipliers, preserving the bound exactly.

**Restore the backup count.** Rightward exploration can only append to this backup array; it never changes the prefix present on entry. Both `shadow_play_right` and `shadow_play_right_small` therefore restore its count instead of copying the backup array out and back. The sorted effective array still has its own snapshot and restoration. The unused 16-byte backup field is removed from `MoveGen` (default rack size).

## Performance

Against `main` at `fb0fc7f7`: M4 Mac mini, CSW24, four workers, WMP/WIT enabled, production static-trained PGO, `simbench` 0.5 s/turn and two plies. Seeds 42/24301/1552 × six interleaved rounds, **18 pairs per configuration**, balanced AB/BA order within each seed. Results are geometric means of paired throughput ratios, with seed-stratified bootstrap 95% CIs (20,000 resamples).

|---|---:|---:|
| **PGO, RIT on** | **+1.28%** | [+1.04%, +1.52%] |
| PGO, RIT off | **+0.65%** | [+0.29%, +0.98%] |

Both binaries were independently trained with `make release PGO_TRAIN_THREADS=4` on the production 16-game static workload, then held fixed. Configuration loading is excluded. All planned pairs were retained; all three seed averages are positive in each configuration. Paired-t intervals also exclude zero: [+0.99%, +1.57%] RIT on and [+0.22%, +1.08%] RIT off. Host load was recorded, with no thermal or performance warnings. These intervals describe run variation on this host with these binaries, not retraining or cross-hardware variation.

Supplementary `no_pgo_release` comparison: same workload and four workers, three seeds × two balanced rounds, **six pairs per configuration**. All 12 pairs are positive. This smaller sample uses paired-t 95% intervals:

|---|---:|---:|
| non-PGO, RIT on | +1.98% | [+1.24%, +2.73%] |
| non-PGO, RIT off | +1.67% | [+0.79%, +2.55%] |

The sampled path is `generate_moves` → `gen_shadow` → `shadow_by_orientation` → `shadow_play_for_anchor` → rightward exploration. Separate 20-second macOS `sample` captures used `BUILD=profile`, seed 42, four workers, 2 s/turn, and WMP/WIT/RIT on.

| function | baseline self samples | PR self samples |
|---|---:|---:|
| `insert_unrestricted_multipliers` | 3,699 | inlined into callers |
| `shadow_play_right` | 5,071 | 7,223 |
| `shadow_play_for_anchor` | 4,732 | 5,631 |

The smaller insertion helper has no standalone symbol in the PR's profile binary, so its samples move into callers. **These counts locate the affected functions; they are not per-function speedup estimates.** The profile build differs from production PGO. In the PGO binaries the insertion helper shrinks from 444 to 352 bytes, `shadow_play_right` grows from 5,032 to 5,136 bytes, and `shadow_play_right_small` shrinks from 1,660 to 1,620 bytes. The throughput measurements capture the net result.

Separate single-worker instrumentation of 16 CSW24 static games records **115,026 unrestricted insertions**, **16,224 backup-element shifts avoided**, and **26,486 backup snapshot/restore pairs eliminated**. There are 26,035 rebuilds, of which 413 encounter non-descending backup entries. Rebuilding may now require more effective-array sorting; it still produces the same sorted multiset. Instrumentation is absent from the timing binaries.

## Correctness

- **New default-suite regression:** `test_shadow_multiplier_rebuild` compares the best score from regular and small-move shadow paths with exhaustive `ALL_SMALL` generation, which bypasses shadow bounds. Fixtures cover letter premiums, cross words, multiple word premiums, blanks, and repeated generator reuse.
- **Mutation checks:** corrupting the saved column trips a bound assertion while the new test runs; removing count restoration triggers an out-of-bounds UBSan failure. The harness uses `UBSAN_OPTIONS=halt_on_error=1`; the correct implementation passes.
- **Full recursive oracle:** 32 seeded CSW24 games, 712 positions and 967,104 complete moves per RIT/WIT combination on the PGO candidate. Move identity, score, and equity agree in all four combinations.
- **Exact shadow bounds:** baseline and PR agree on all 116,902 anchors with RIT on and 124,595 with RIT off, WIT enabled, across the same 32 games. Anchor metadata and both score/equity bounds match.
- **Multiplier invariant:** an instrumented build independently reconstructs and sorts the effective multiset after every insertion; static autoplay and the movegen/shadow suites pass.
- Final `movegen`, `shadow`, `wmg`, and `witdiff` suites pass under ASan/UBSan, including 32 games/723 positions compared before and after copy/undo. Formatting, circular dependencies, changed-file cppcheck and host-adapted clang-tidy checks pass; all 16 CI checks pass on `3deb2569`.

## What can build on this

The effective list still needs sorted insertion. Further changes to its representation should measure both common insertions and rebuilds, preserve the exact bound, and retain restoration of the sorted array after rightward exploration. No endgame throughput gain is claimed from the smaller small-move routine alone.


src/impl/move_gen.c
```diff
@@ -1977,13 +2023,9 @@ static inline void maybe_recalculate_effective_multipliers(MoveGen *gen) {
   const int original_num_unrestricted_multipliers =
       gen->num_unrestricted_multipliers;
   gen->num_unrestricted_multipliers = 0;
-  // We insert the columns with highest cross wordletter multipliers first
-  // so the list will mostly sort in the the order it is traversed,
-  // minimizing the number of swaps.
   for (int i = 0; i < original_num_unrestricted_multipliers; i++) {
```
olaugh, line 2026: One consequence worth knowing (no change needed): the deleted comment noted that inserting high cross-word multipliers first kept this rebuild's insertion sort nearly-swap-free. With the backup array now unsorted, this loop's insert can degrade to O(m²) swaps per rebuild — but m ≤ 7 and rebuilds only happen on main-word multiplier changes, so the per-square sorted insert you removed dominates. The +2.9% median confirms the trade.

olaugh (commented):
Reviewed for correctness: the cross-word component array's order is genuinely unobserved — `maybe_recalculate_effective_multipliers` visits every retained entry and does a sorted insert into the effective array regardless of input order, and the counter increment stays in `insert_unrestricted_effective_letter_multiplier`, so append + counter bump is equivalent to the old sorted insert + bump. Renames make the invariant much clearer, and the equivalence tests are appreciated. LGTM.

jvc56 (approved):

PR #621, opened 2026-07-23
Emit only touched WMP shadow anchors

olaugh:
## Summary

#620 made WMP shadow-anchor reset sparse by tracking the slots touched during
`shadow_play_for_anchor`. Emission still scanned all
`MAX_WMP_MOVE_GEN_ANCHORS` slots (64 on the standard board) after every board
anchor, even though a typical pass touches no more than about eight.

This PR replaces the touched-slot list with a compact mask and walks only its
set bits for both reset and emission.

The mask is enumerated by word and then by least-significant set bit, preserving
the old full-array scan's ascending slot order. That matters for deterministic
tie ordering. The mask is an array of 64-bit words rather than a single word,
so BOARD_DIM=21 builds remain supported.

## Performance

Standardized simbench, relative to #620:

- `no_pgo_release`
- 8 threads
- RIT on
- WIT on
- `SIMBENCH_MI=100000`
- 2 seconds per turn
- six interleaved paired rounds with alternating order

| Pair | #620 | This PR | Paired change |
|---:|---:|---:|---:|
| 1 | 21,822 it/s | 22,352 it/s | +2.43% |
| 2 | 21,424 it/s | 22,226 it/s | +3.74% |
| 3 | 21,224 it/s | 21,148 it/s | -0.36% |
| 4 | 21,287 it/s | 21,920 it/s | +2.97% |
| 5 | 21,263 it/s | 22,121 it/s | +4.04% |
| 6 | 21,258 it/s | 22,126 it/s | +4.08% |
| **Median** |  |  | **+3.36%** |

Five of six pairs improved. The one small negative pair was followed by three
additional gains around 3-4%.

## Correctness

- Slot traversal remains ascending, matching the previous `for (i = 0; i <
  MAX_WMP_MOVE_GEN_ANCHORS; i++)` output order exactly.
- A targeted test touches slots in descending order, verifies ascending
  emission, resets, and verifies both mask and anchor sentinels are cleared.
- The mask spans multiple words for super-board builds.

Passed locally:

- `wmg shadow movegen wit sim`
- `ap_wmp`: 1,000 game pairs for each of 12 installed lexica (24,000 games),
  comparing WMP movegen with legacy recursive movegen
- full `BOARD_DIM=21` no-PGO test run
- formatter and `git diff --check`


src/impl/wmp_move_gen.h
```diff
@@ -88,16 +85,35 @@ typedef struct WMPMoveGen {
   uint32_t playthrough_positions;
 } WMPMoveGen;
 
+static inline int wmp_move_gen_get_lowest_set_bit(uint64_t bits) {
```
olaugh, line 88: The portable fallback loops forever if `bits == 0`. Both callers guard with `while (touched != 0)`, so it's unreachable today, but an `assert(bits != 0)` here would keep a future caller from turning that into a hang instead of an assertion failure.

olaugh (commented):
Verified the mask protocol preserves the old semantics: bit set ⟺ the slot's `tiles_to_play` sentinel is nonzero (first-touch marking in `wmp_move_gen_maybe_update_anchor` is unchanged), word-then-LSB enumeration preserves the full scan's ascending slot order for deterministic ties, and the emission's cutoff condition is a correct inversion of the old skip condition. The `WMP_ANCHOR_MASK_WORDS` array keeps BOARD_DIM=21 (88 slots) working — good that CI runs both dims. Pushed a small commit adding the two missing direct includes clang-tidy wanted (board_defs.h, equity_defs.h); #623 had already added the same ones upstream, so the merge dedupes cleanly.

PR #623, opened 2026-07-23
Initialize WMP shadow anchor slots once

olaugh:
## Summary

`wmp_move_gen_init` currently seeds every WMP shadow-anchor slot on every
WMP-enabled position:

- 64 slots on the standard board
- 88 slots in a `BOARD_DIM=21` build

That full pass is necessary before first use because the zeroed `Anchor`
representation does not contain the required `EQUITY_MIN_VALUE` and
`BOARD_DIM - 1` sentinels. It is unnecessary after that: #621 already tracks
every modified slot and can reset those slots sparsely.

This PR keeps a lifecycle bit in the calloc-owned, per-thread `MoveGen`.
The first active WMP load seeds every slot; later active loads reset only the
slots touched by the prior position. An inactive load deliberately preserves
both the lifecycle bit and touched masks so a later active load can consume
the outstanding reset.

Standalone `WMPMoveGen` callers pass no lifecycle state and retain the old
always-seed behavior. This avoids reading an indeterminate flag from their
stack storage.

## Performance

Standardized simbench, relative to #622:

- `no_pgo_release`
- 8 threads
- RIT on
- WMP on
- WIT on
- `SIMBENCH_MI=100000`
- 2 seconds per turn
- nine interleaved paired rounds with alternating order

| Pair | #622 | This PR | Paired change |
|---:|---:|---:|---:|
| 1 | 23,424 it/s | 23,208 it/s | -0.92% |
| 2 | 22,861 it/s | 23,099 it/s | +1.04% |
| 3 | 22,769 it/s | 22,894 it/s | +0.55% |
| 4 | 22,683 it/s | 22,844 it/s | +0.71% |
| 5 | 22,647 it/s | 22,768 it/s | +0.53% |
| 6 | 22,616 it/s | 22,786 it/s | +0.75% |
| 7 | 22,673 it/s | 22,670 it/s | -0.01% |
| 8 | 22,568 it/s | 22,690 it/s | +0.54% |
| 9 | 22,531 it/s | 22,660 it/s | +0.57% |
| **Median** |  |  | **+0.55%** |
| **Mean of middle three** |  |  | **+0.55%** |

Seven pairs improved, one was effectively flat, and one regressed. The robust
center is positive but below this protocol's usual ~1% resolution, so this
should be treated as a small directional result rather than a standalone
conclusive speedup.

### Causation check

I also profiled the exact #622 and #623 benchmark binaries rather than relying
only on elapsed throughput:

- arm64 disassembly confirms that #622 executes the 64-slot sentinel loop on
  every active load, while #623 emits the intended lifecycle check and
  touched-mask walk; the 64-slot loop remains only as the first-use fallback;
- matched 30-second macOS `sample` captures put exclusive/top-of-stack
  `gen_load_position` samples at 315 for #622 and 285 for #623 (-9.5% in the
  containing function);
- temporary work counters over one standard simbench game observed 4,326,296
  active WMP loads. #622's unconditional loop would initialize 276,882,944
  slots; #623 initialized or reset 9,948,055 slots: 64.00 down to 2.30 slots
  per load, a 96.41% reduction in the targeted operation.

The work-count and disassembly evidence directly validate that the intended
work disappeared. The 30-sample drop in `gen_load_position` is directionally
consistent but too small to be independently conclusive. In particular, none
of this upgrades the +0.55% end-to-end result beyond “small and likely real”:
the benchmark remains below the protocol's usual ~1% noise floor.

## Correctness

The main failure mode is clearing a touched mask without first resetting the
slots it identifies. That would leave stale maxima whose next use might not
mark the slot touched, potentially underestimating or misordering an anchor.

A targeted lifecycle test verifies:

- an inactive first call does not claim initialization;
- first active use seeds every required sentinel;
- the highest-index slot is tracked and reset;
- active → inactive preserves the outstanding tracker;
- the following active initialization consumes the tracker and restores every
  sentinel.

Passed locally:

- `wmg shadow movegen wit sim`
- `ap_wmp`: 1,000 game pairs for each of 12 installed lexica (24,000 games),
  comparing WMP movegen with legacy recursive movegen
- full `BOARD_DIM=21` no-PGO test run
- formatter and `git diff --check`


src/impl/wmp_move_gen.h
```diff
@@ -129,22 +133,29 @@ static inline void wmp_move_gen_init(WMPMoveGen *wmp_move_gen,
...
-    wmp_move_gen->anchors[i].leftmost_start_col = BOARD_DIM - 1;
-    wmp_move_gen->anchors[i].tiles_to_play = 0;
-  }
-  memset(wmp_move_gen->touched_anchor_masks, 0,
-         sizeof(wmp_move_gen->touched_anchor_masks));
+  if (anchor_slots_initialized != NULL && *anchor_slots_initialized) {
+    // The prior position can leave its final board anchor touched. Reset
+    // those slots before reusing the tracker; clearing the masks alone would
+    // lose the only record that their values are stale.
+    wmp_move_gen_reset_anchors(wmp_move_gen);
```
olaugh, line 140: Correction to my initial read: the `wmp == NULL` early return above means an inactive load never reaches this branch at all — the lifecycle bit and touched masks are genuinely preserved untouched, exactly as the PR description says, and the outstanding reset is consumed here on the next *active* load. The eager-looking reset is therefore correctly deferred across inactive loads. No change requested.

olaugh (commented):
The lifecycle contract is sound: the flag lives in the calloc-owned per-thread `MoveGen` (so first use is reliably false), standalone `WMPMoveGen` users pass NULL and keep always-seed behavior, and the `wmp == NULL` early return preserves both the flag and the touched masks across inactive loads so the next active load consumes the outstanding reset — the case that would silently corrupt bounds if missed. Since the sentinels don't depend on which WMP is loaded, lexicon swaps are safe too. Lifecycle test coverage is exactly what this kind of stateful optimization needs. LGTM.

PR #624, opened 2026-07-23
Consolidate simulation rollout statistics updates

olaugh:
## Summary

Stacked on #619 (`wmp-playthrough-first`).

- add compile-time-gated, tagged lock profiling (`LOCK_PROFILE=1`) with no production instrumentation when disabled
- collect each rollout's display/statistics updates locally and commit them under one stats lock
- keep seed generation separately synchronized
- publish floating statistics in arm-local ordinal order through a bounded reorder buffer
- preserve partial rollouts, heat maps, zero/nonzero spread utility, and immediate BAI updates

This intentionally does not change move generation or BAI sampling semantics.

## Why

Tagged attribution showed that most `__ulock_wait` samples were legitimate joins rather than useful contention. The actionable site was `SimmedPlay` statistics/seed synchronization.

At 18 threads, the candidate reduced:

- stats/seed acquisitions by 66.1%
- contended acquisitions by 82.1%
- measured stats blocking wait from 616 ms to 146 ms (76.4%)

The normal macOS profile agreed: `__psynch_mutexwait` samples fell 11.4% and `pthread_mutex_lock` samples fell 27.9%. `__ulock_wait` remained roughly flat because joins dominate it.

## Performance

M5 Max on AC power in High Power mode, `no_pgo_release`, one autoplay game, 18 cooperating in-game-parallel threads, CSW24, two-ply simulations, RIT/WIT/WMP enabled, two seconds per turn, seed 42, and alternating baseline/candidate order. One warm-up of each binary was excluded before nine measured pairs.

| Pair | #619 baseline iter/s | #624 candidate iter/s | Change |
|---:|---:|---:|---:|
| 1 | 271,816 | 294,015 | +8.167% |
| 2 | 297,382 | 298,835 | +0.489% |
| 3 | 295,706 | 301,973 | +2.119% |
| 4 | 296,283 | 299,705 | +1.155% |
| 5 | 296,189 | 301,072 | +1.649% |
| 6 | 294,591 | 301,748 | +2.429% |
| 7 | 295,268 | 301,580 | +2.138% |
| 8 | 297,797 | 302,503 | +1.580% |
| 9 | 296,433 | 301,637 | +1.756% |

- paired median: **+1.756%**
- ratio of medians: **+1.820%** (296,189 to 301,580 iter/s)
- all 9/9 pairs favored the candidate
- excluding the unusually slow first baseline, the median of the remaining eight paired changes is **+1.702%**

This confirms the earlier non-High-Power result (**+1.543% median; 8/9 candidate wins**) and clears the project's 1% noise threshold.

### Benchmark-harness normalization

#624 makes the standardized simbench command explicit by adding `-mtmode igp`; the #619 benchmark harness predates that addition and otherwise uses its older default mode. For this A/B, the temporary #619 worktree received the same test-only `-mtmode igp` argument so both binaries ran the identical single-game workload. Product code in the baseline was unchanged. An initial run using the two different harness defaults produced a spurious order-of-magnitude gap and was discarded.

Reproduction settings:

```sh
SIMBENCH_THREADS=18 \
SIMBENCH_TLIM=2 \
SIMBENCH_RIT=true \
SIMBENCH_WIT=true \
SIMBENCH_WMP=true \
./bin/magpie_test simbench
```

Build both the #619 head and this branch with:

```sh
make magpie_test BUILD=no_pgo_release LOCK_PROFILE=0
```

When benchmarking the #619 commit directly, also normalize its test-only simbench command to include `-mtmode igp`.

## Correctness

Previously passed during the investigation:

- all non-network tests under ASan/UBSan
- focused BAI and simulation tests
- focused ThreadSanitizer suite
- exact raw display-stat double bit patterns across 1-11 threads under round-robin scheduling
- heat maps enabled and disabled
- spread utility weight zero and nonzero
- early termination with fewer completed plies than requested

Re-run on the current #624 head before the High Power benchmark:

- `magpie_test bai`: passed
- `magpie_test sim`: passed
- `magpie_test simexact`: passed, including exact results across 1-11 threads

The existing top-two BAI result is already not bit-identical across thread counts because physical completion order changes per-arm sample counts; the untouched #619 baseline has the same behavior. This change neither worsens nor fixes that pre-existing limitation.


src/ent/stats.c
```diff
@@ -93,6 +93,23 @@ double stat_get_margin_of_error(const Stat *stat, double zval) {
...
 
+static bool double_bits_equal(double x, double y) {
+  uint64_t x_bits;
+  uint64_t y_bits;
+  memcpy(&x_bits, &x, sizeof(x_bits));
+  memcpy(&y_bits, &y, sizeof(y_bits));
+  return x_bits == y_bits;
+}
+
+bool stat_is_bit_identical(const Stat *a, const Stat *b) {
```
olaugh, line 104: `stat_is_bit_identical` is only called from `test/test_util.c`, and CLAUDE.md's rule is that test-only code lives in `test/`. It's here because `Stat` is opaque and there are no getters for `sum_of_mean_differences_squared` or `mean_is_estimated`. Two options: add those two getters and move the comparison into test_util, or keep it here with a comment marking it test-support. Worth deciding before this leaves draft.

src/ent/sim_results.c
```diff
@@ -27,6 +30,19 @@ typedef struct PlyInfo {
   uint64_t ply_info_counts[NUM_PLY_INFO_COUNT_TYPES];
 } PlyInfo;
 
+typedef struct PendingRolloutStats {
+  uint64_t ordinal;
+  int num_completed_plies;
+  Move ply_moves[MAX_PLIES];
```
olaugh, line 36: Sizing note: `ply_moves[MAX_PLIES]` makes each pending slot ~1KB, and there are `num_threads` slots per simmed play — e.g. 40 plays × 18 threads ≈ 700KB. Not a problem today, but since `num_plies` is known at reset time, the slots could be sized to the actual ply count if this ever grows (deep-ply heat-map sims are the case that would).

src/util/lock_profile.c
```diff
@@ -0,0 +1,234 @@
...
+  if (pthread_mutex_unlock(&lock_profile_registry_mutex)) {
+    log_fatal("lock profile report unlock failed");
+  }
+
+  uint64_t all_wait_ns = 0;
+  for (int site = 0; site < NUM_LOCK_PROFILE_SITES; site++) {
+    all_wait_ns +=
+        totals[site].blocking_wait_ns + totals[site].condition_wait_ns;
+  }
+  printf("LOCK_PROFILE hold_sample_period=%d registered_threads=%d\n",
```
olaugh, line 187: CLAUDE.md bans printf in production code in favor of the status print mechanism. This is compile-gated diagnostics that never ships in a normal build, which seems like a reasonable exception — but it's worth a one-line comment saying so, or routing through `thread_control_print` if that's practical at report time, so the next reader doesn't flag it.

olaugh (commented):
Reviewed the reorder-buffer protocol closely — it's sound. Deadlock-freedom argument: each worker holds at most one uncommitted ordinal (seed drawn at rollout start, committed unconditionally at the end with no early return in between), and the worker holding the arm's next ordinal never enters the wait loop, so progress is always possible; cross-arm waits can't cycle because a blocked worker has no other outstanding ordinal. The `capacity - 1` wait threshold guarantees `find_empty_pending_slot` always succeeds, making the `log_fatal` correct-but-unreachable defense. The bit-identical-across-thread-counts property (with `-sr rr`) plus the tests enforcing it is a genuinely valuable addition beyond the perf win.

I pushed two commits fixing the CI failures: the analyzer null-deref pattern at the pending-slot insert, bitwise double comparison in `stat_is_bit_identical` (memcmp of doubles trips cert-flp37-c), missing/unused direct includes, and format.py-compatible conditional includes in lock_profile.c (plus removing an accidentally staged gitlink — my fault, not yours).

olaugh:
Closing this PR because we cannot currently vouch for the end-to-end speedup. The earlier performance claims in the description should not be treated as established against current main.

A new M5 Max run verified 18 physical/logical cores and independently trained native production-PGO baseline and candidate binaries on this machine. Baseline was main `39d16e4603d61b61181fa85379266bdd1cce2ce6`, including merged #612. The candidate was the prepared, revised `waiters.patch` on that baseline—not the current GitHub PR head. It includes batched statistics, borrowed immediate-commit moves, owned deferred moves, corrected duplicate-snapshot initialization, and broadcasts only when waiters exist. Patch SHA-256: `a62ae3902f98aca40d6e4979885aec196fd1eeaaae74027a1823cfb05449356f`.

Independent confirmation used CSW24, WMP/WIT/RIT enabled, two plies, 0.5 seconds/turn, explicit IGP, three seeds, and balanced AB/BA ordering. All 54 planned pairs were completed and retained, with 18 pairs per worker count:

| Workers | Geometric mean throughput gain | Paired 95% interval |
|---|---:|---:|
| 1 | +0.02% | −0.87% to +0.91% |
| 8 | +0.03% | −1.27% to +1.35% |
| 18 | +0.33% | −1.76% to +2.47% |

There is no established throughput improvement or serial regression. Both variants scaled similarly: about 7.48× serial throughput at 8 workers and 13.9–14.0× at 18. Intermittent iCloud/FileProvider activity limits precision; no competing compiler was captured in confirmation health snapshots. The preceding screen overlapped other work and is not the primary result. No unfavorable pairs were discarded. Older M4 measurements predate #612 and are not current-main evidence.

Correctness checks passed separately: ASan/UBSan and ThreadSanitizer `sim` and `simexact`, plus fresh fixed-budget raw-statistics comparisons against serial baseline. Candidate exports at 1/8/18 workers matched all 4,128 records bit-for-bit per worker count, including floating accumulator bits, counts and heat maps.

Separate profiles and instrumented diagnostics confirm about 80% fewer statistics-lock acquisitions (roughly five per rollout to one, plus a seed lock in both). At 18 workers, seed-ordered publication produced 879 condition waits across 3.91 million diagnostic rollouts, totaling 0.514 summed thread-seconds. Move generation still dominated the sampled rollout path. The mechanism reduces synchronization costs, but that did not establish an end-to-end gain in the uninstrumented PGO confirmation.

The complete report, original bundle, exact patches, scripts, native profiles/binaries, raw logs, health snapshots and analyses are preserved locally on maxim.local at `/Users/john/sources/jul24-leavegen/pr624-m5-evidence-20260907/` (`REPORT.md` is the index).

This is a performance-evidence decision, not a correctness failure. Any future proposal should establish a reproducible benefit against then-current main before reviving the speedup claim. Closing without merging.

PR #625, opened 2026-07-24
Use root PVS for top-one endgame searches

olaugh:
## What this does

Use principal-variation search at the root when only the best endgame move is requested. After the first move, later moves receive a null-window probe and are searched again with a wider window when they improve alpha. Single-thread top-one searches also use an aspiration window around the previous completed depth's value, widening and re-searching failures.

Multi-PV, per-ply/per-root reporting, and historical actual-move extraction retain their existing search paths. Actual-move callers are excluded from the new single-thread aspiration path so a low-valued played move is not reported as a bound.

## Performance

**+6.44% geometric-mean speedup; +2.54% paired median.** Measured against main `64019d5d` on an Apple M4 Mac mini (10 cores, 16 GiB), macOS 15.7.4, Apple Clang 17. Both sources received independent production PGO training from 16 static CSW24 games with four training workers. WMP/RIT/WIT were enabled with identical CSW24 assets.

The primary workload contains **64 fresh, unfiltered bag-empty positions**, generated by static-equity self-play with seed `420000103`. Four paired rounds alternate whole-arm order. Each solve uses one thread, depth four, seed 42, and a fresh owned TT (256 MiB minimum). Solver allocation/reset is timed; configuration/data loading and destruction are outside the timer. All 512 timed solves are retained.

| Measurement | Result |
|---|---:|
| Geometric mean of per-position median speed ratios | **+6.44%** |
| Nominal paired 95% interval | **[+2.98%, +10.02%]** |
| Median across positions | **+2.54%** |
| Positions favoring this change | **48/64** |
| Per-position range | −6.55% to +103.27% |
| Total-time speedup across all four rounds | +8.13% |
| Nodes per full corpus | 38,634,528 → 36,458,119 (**−5.63%**) |

The interval is a Student-t interval over 64 paired log ratios, using each position's median ratio across rounds. Repeated rounds are not treated as independent positions. Background iCloud activity occupied roughly one core in boundary snapshots; continuous thermal/frequency telemetry was not collected. These figures describe this host and pair of trained binaries, not cross-hardware or repeated-training uncertainty.

Secondary controls are descriptive single paired runs:

- Eight-thread depth-four corpus: 73.77 → 62.26 seconds, with all 64 values matching. Parallel scheduling adds variability, so this is not the headline estimate.
- Top-three depth-three control: 115.44 → 115.84 seconds, with identical nodes, values, and moves. The optimization is disabled for this path.
- Sixty-four fresh small positions (at most six rack tiles): values agree at both 12 and 24 plies; geometric speedups are +1.60% and +1.74%. Stability across these depths is not a terminal-proof claim.
- Eight-thread, 100 ms deadline control: the candidate completes a greater depth on 30/64 positions, main on none. This measures progress, not decision quality against an oracle.

Primary timing/profiles use engine commit `fc7143b1`. Final head `fc6dff1d` changes only the benchmark driver to enforce timed-control deadlines and keep I/O outside assertions. The corrected deadline control was measured separately; its raw results are distinct from the fixed-depth campaign.

Six captures sample the frozen production-PGO executables for six seconds each, after one second of startup, using three corpus rotations. Exclusive shares among named engine leaf samples were:

| Function/path | Main | This PR |
|---|---:|---:|
| `recursive_gen_small` + `go_on_small` | 34.44% | 33.54% |
| `shadow_play_for_anchor_small`, `shadow_play_right_small`, `shadow_by_orientation_small` | 28.64% | 28.55% |
| `game_gen_cross_set` + `game_gen_cross_set_tracked` | 10.06% | 10.50% |
| `abdada_negamax` + `negamax_greedy_leaf_playout` self | 2.20% | 2.03% |

`abdada_negamax` and `iterative_deepening` change the search windows; the benefit comes from visiting fewer descendants. Small-move generation, shadow, and cross-set updates dominate the remaining search work. Their implementations are unchanged. These normalized sample shares are not per-function speedups or call counts; fixed capture windows can cover different roots. System/wait samples and functions with fewer than five samples per capture are excluded from this table.

## Correctness

- All **256 primary paired values** agree; all 64 parallel values agree.
- Different selected moves occur at two primary positions and seven parallel positions. Sixteen independent main-branch actual-move solves verify every distinct selected move is tied with the best value at the measured depth.
- Optimized and ASan/UBSan endgame suites pass, including existing reporting, multi-PV, interruption, and dual-lexicon coverage.
- A new default-suite regression compares a historical pass with an independently solved child position. Removing the actual-move exclusions makes that regression fail at the expected value assertion.
- Full cppcheck 2.17.1, formatting, dependency-cycle checks, and `git diff --check` pass. Full platform CI is attached to the PR.

The on-demand `egrootbench` driver supports reproducible corpus generation and fixed-depth/deadline controls. Use the same driver on both sources, fresh PGO profiles, and separate TTs.

## What can build on this

Root ordering and aspiration-width tuning may reduce re-searches further. Any follow-up should retain the actual-move/reporting exclusions and measure nodes, elapsed time, and completed values independently.

### DIFF
--- a/src/impl/endgame.c
+++ b/src/impl/endgame.c
@@ -2473,6 +2473,14 @@ int32_t abdada_negamax(EndgameCtxWorker *worker, uint64_t node_key, int depth,
   }
   const int multi_pv_k = worker->solver->num_top_moves;
   const bool multi_pv = is_root && multi_pv_k > 1;
+  // Multi-PV and reporting callbacks expose values for multiple root moves, so
+  // those callers need full-width searches at the root. A top-1 solve without
+  // either callback only needs the best move: later roots can use the normal
+  // PVS null-window probe and are re-searched at full width only if they
+  // improve alpha.
+  const bool exact_all_root_moves =
+      is_root && (multi_pv || worker->solver->per_ply_callback != NULL ||
+                  worker->solver->per_root_move_callback != NULL);
   // Sized for the live multi-PV leaderboard breadth (up to
   // MAX_ENDGAME_DISPLAY_PVS root moves), not the per-line depth. num_top_moves
   // is clamped to MAX_ENDGAME_DISPLAY_PVS in endgame_ctx_reset so topk_insert
@@ -2631,7 +2639,8 @@ int32_t abdada_negamax(EndgameCtxWorker *worker, uint64_t node_key, int depth,
       // ID iteration. This gives accurate values for all root moves (needed for
       // multi-PV) while still benefiting from narrow windows.
       const bool use_root_aspiration =
-          is_root && depth >= 2 && worker->solver->iterative_deepening_optim &&
+          exact_all_root_moves && depth >= 2 &&
+          worker->solver->iterative_deepening_optim &&
           !worker->solver->first_win_optim &&
           !worker->solver->initial_window_optim;
 
@@ -2670,7 +2679,8 @@ int32_t abdada_negamax(EndgameCtxWorker *worker, uint64_t node_key, int depth,
             break;
           }
         }
-      } else if (idx == 0 || !worker->solver->negascout_optim || is_root) {
+      } else if (idx == 0 || !worker->solver->negascout_optim ||
+                 exact_all_root_moves) {
         value =
             abdada_negamax(worker, child_key, depth - 1, -beta, -alpha,
                            &child_pv, pv_node, child_exclusive, opp_stuck_frac);
@@ -2947,7 +2957,15 @@ void iterative_deepening(EndgameCtxWorker *worker, int plies) {
   // value hasn't changed — result is stable and we can bank the remaining time.
   // INT32_MIN is the sentinel meaning "not yet crossed the soft limit".
   int32_t soft_limit_pv_value = INT32_MIN;
-  bool use_aspiration = (worker->solver->threads > 1);
+  // Single-thread multi-PV/per-root reporting already gives every root move a
+  // private aspiration window in abdada_negamax.  Top-1 searches use PVS at
+  // the root, so a narrow window around the preceding IDS value can prune the
+  // whole tree without sacrificing an exact result (failures are widened and
+  // re-searched below).
+  bool use_aspiration = worker->solver->threads > 1 ||
+                        (worker->solver->num_top_moves == 1 &&
+                         worker->solver->per_ply_callback == NULL &&
+                         worker->solver->per_root_move_callback == NULL);
 
   if (worker->solver->first_win_optim) {
     // search a very small window centered around 0; we're just trying to find
--- a/src/impl/endgame.c
+++ b/src/impl/endgame.c
@@ -310,6 +310,14 @@ struct EndgameCtxWorker {
 #define MIN(a, b) ((a) < (b) ? (a) : (b))
 #endif
 
+// These consumers need root values beyond the best move and must share the
+// same search-mode decision in root PVS and outer aspiration setup.
+static bool solver_needs_exact_root_values(const EndgameCtx *solver) {
+  return solver->num_top_moves > 1 || solver->actual_move != NULL ||
+         solver->per_ply_callback != NULL ||
+         solver->per_root_move_callback != NULL;
+}
+
 // Insert a value into a sorted (descending) top-K array.
 // Returns the Kth-best value (or -LARGE_VALUE if fewer than K values stored).
 static inline int32_t topk_insert(int32_t *topk, int *n, int k, int32_t val) {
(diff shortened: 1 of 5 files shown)


olaugh (commented):
Traced the correctness properties this rests on and they all hold:

- The scout path is textbook PVS: probe at `(-alpha-1, -alpha)` with `pv_node=false`, re-search full-width (non-exclusive, as the comment says) only when `alpha < -value < beta`. ABDADA deferral (`ON_EVALUATION`) flows through the probe unchanged.
- Reporting safety in top-1 mode: fail-low probes publish upper bounds via `publish_live_top_k_pv`, but a K=1 leaderboard insert can never let a bound (≤ current alpha ≤ best) displace the exact, re-searched best — so the live display stays truthful.
- `extract_multi_pvs`, whose header comment requires estimated values to be "the actual negamax return value", is gated on `num_top_moves > 1`, which forces `exact_all_root_moves` — so the one consumer that needs exact per-root values can never see PVS bounds. Callback consumers (TUI streaming, per-root reporting) likewise force the exact path.
- Root fail-high/low under the newly-enabled single-thread aspiration lands in the existing widen-and-re-search loop.

The empirical validation is the right kind: re-solving all 19 disagreements to exact terminal depth with the same oracle and getting 13-better/6-tied/0-worse directly tests that the extra depth bought better decisions, not just different ones. CI is green including the fixed-depth endgame tests, which now exercise the root-PVS path and still produce exact results.

One suggestion beyond the inline comments: a cheap determinism test would lock this in — solve a couple of fixtures to fixed depth top-1 twice, once callback-free (PVS root) and once with a dummy per-root callback (exact-all root), and assert identical best move and value. That's the invariant the whole PR hangs on, and it would catch any future drift in the gating predicate.

src/impl/endgame.c
```diff
@@ -2473,6 +2473,14 @@ int32_t abdada_negamax(EndgameCtxWorker *worker, uint64_t node_key, int depth,
   }
   const int multi_pv_k = worker->solver->num_top_moves;
   const bool multi_pv = is_root && multi_pv_k > 1;
+  // Multi-PV and reporting callbacks expose values for multiple root moves, so
+  // those callers need full-width searches at the root. A top-1 solve without
+  // either callback only needs the best move: later roots can use the normal
+  // PVS null-window probe and are re-searched at full width only if they
+  // improve alpha.
+  const bool exact_all_root_moves =
```
olaugh, line 2481: This predicate and its inverse in `iterative_deepening` (the new `use_aspiration` condition, ~line 2965) encode the same mode decision in two places. If a future caller adds another root-value consumer (a third callback, say) and only updates one site, the modes desynchronize silently — exact-all root searches under a narrow root window, or PVS probes feeding a consumer that expects exact values. A tiny shared helper like `solver_needs_exact_root_values(solver)` would make that drift impossible.

src/impl/endgame.c
```diff
@@ -2670,7 +2679,8 @@ int32_t abdada_negamax(EndgameCtxWorker *worker, uint64_t node_key, int depth,
             break;
           }
         }
-      } else if (idx == 0 || !worker->solver->negascout_optim || is_root) {
+      } else if (idx == 0 || !worker->solver->negascout_optim ||
+                 exact_all_root_moves) {
```
olaugh, line 2683: Anchoring here since the estimated-value line isn't in the diff: with this branch taken for later top-1 roots, `small_move_set_estimated_value` (~line 2772) now stores an upper bound (probe fail-low) rather than an exact value for most root moves. That's the standard PVS trade and only affects next-iteration move ordering plus aspiration centers — but the comment at ~line 3377 ("The estimated_value for each root move is the actual negamax return value") is now only true in exact-all mode. Since that comment is load-bearing for `extract_multi_pvs` (safely gated on `num_top_moves > 1`), it's worth qualifying it so the invariant reads correctly.

src/impl/endgame.c
olaugh: Fixed in 19169394. Both root PVS and outer aspiration setup now use solver_needs_exact_root_values(), including multi-PV, actual-move extraction, and both reporting callbacks. The existing single-thread top-one gate is preserved. The endgame suite passes under ASan/UBSan, including the actual-pass regression and callback/multi-PV coverage.

src/impl/endgame.c
olaugh: Fixed in 19169394. The comment now explicitly says that top-one PVS fail-low estimates are upper bounds used for next-iteration ordering. It also explains that extract_multi_pvs is gated on num_top_moves > 1, excludes the top-one PVS path, and still handles bounds or mixed depths from multi-PV cutoffs and interruptions through sorting and full-window resolution. The ASan/UBSan endgame suite passes.

jvc56 (approved):

PR #626, opened 2026-07-24
Score all endgame outplays analytically

olaugh:
## Summary

- score every rack-emptying endgame move exactly in its parent node
- skip SmallMove conversion, game and Zobrist mutation, terminal child negamax entry, and unplay
- replace the narrow single-legal-move, one-tile shortcut with one path that also covers root, multi-PV, callback, and deeper-node outplays
- add a multiple-root-outplay regression that verifies every reported value and confirms terminal child nodes are not entered

## Correctness

An outplay in a bag-empty game has the exact current-player value:

`current spread + move score + 2 * opponent rack value`

The search keeps the existing negamax sign convention by storing the negative of that value in the child-value variable. Pass is explicitly excluded. The normal parent PV, alpha/beta, multi-PV, callback, progress, and TT-store paths remain responsible for consuming the result.

Validation completed:

- `magpie_test endgameoutplay`
- `magpie_test endgame`
- `magpie_test multipv`
- sanitizer build with strict warnings
- exact fixed-depth values matched for every benchmark position and round

## Performance assessment

This is a smaller result than the removed node count initially suggested.

`no_pgo_release`, CSW24 RIT on, 20 positions, 4 ply, five interleaved paired rounds:

| Threads | Exact-value disagreements | Node change | Elapsed result |
|---:|---:|---:|---:|
| 1 | 0 | **-7.12%** exactly every round | middle-three mean **+0.02%**; flat |
| 10 | 0 | -6.5% to -7.8% by round | middle-three mean **+0.98%**; no win |

The single-thread paired-position median was -0.23%, below the approximately 1% noise floor. The terminal child calls are real nodes but almost costless, so removing them does not produce a measurable wall-clock speedup on this battery. This draft deliberately makes no throughput claim; it is primarily a correct simplification and a possible building block for later terminal-search work.

## Review focus

The reviewer should probe negamax sign handling and all reporting paths, especially multiple root outplays. The new regression checks both: callback values equal `move score + opponent rack bonus`, and at depth one the total node count is exactly root plus the non-outplay children.

### DIFF
--- a/src/impl/endgame.c
+++ b/src/impl/endgame.c
@@ -2426,42 +2426,6 @@ int32_t abdada_negamax(EndgameCtxWorker *worker, uint64_t node_key, int depth,
   size_t arena_offset =
       worker->small_move_arena->size - (sizeof(SmallMove) * nplays);
 
-  // Analytical leaf shortcut for single-tile outplay.
-  // When stm has exactly 1 tile and the only legal move is a non-pass, it
-  // plays that tile and ends the game immediately. Skip the board mutation
-  // cycle (small_move_to_move + play_move + child recursion + unplay).
-  if (arena_alloced && nplays == 1) {
-    const SmallMove *only_sm =
-        (const SmallMove *)(worker->small_move_arena->memory + arena_offset);
-    const Rack *stm_rack_a = player_get_rack(player_on_turn);
-    if (!small_move_is_pass(only_sm) &&
-        rack_get_total_letters(stm_rack_a) == 1) {
-      const LetterDistribution *ld = game_get_ld(worker->game_copy);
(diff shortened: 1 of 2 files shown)


olaugh:
Closing without merge after the High Power 18-thread precision retest. Across 21 interleaved pairs of 100 exact CSW21 endgames, all 2,100 cross-build values matched, but the paired median wall-clock result was -0.490% for the candidate and the geometric-mean result was -0.996% (95% CI -3.126% to +1.181%). The candidate searched 10.220% fewer nodes at the median, but that did not translate into a measurable speedup; it won only 10/21 pairs. The removed terminal nodes are real but too cheap, and the result remains within parallel-search variance, so there is no defensible throughput claim.

PR #627, opened 2026-07-24
Compile TT statistics out of release search

olaugh:
## Summary

- compile the four transposition-table statistics increments out of `NDEBUG` engine translation units
- retain counters, initialization, and reset behavior for a stable structure layout
- keep statistics active in development and diagnostic/test translation units
- use relaxed ordering for diagnostic increments, which do not synchronize search state

## Why

Every TT lookup previously incremented `lookups`, every hit incremented `hits`, every type-2 collision incremented `t2_collisions`, and every store incremented `created`. These values are diagnostic only, but the atomics share a cache line across search threads.

Optimized native builds define `NDEBUG`, so their hot lookup/store paths now contain no counter RMW. Optimized test source files are deliberately compiled without `NDEBUG`, preserving the existing counter unit test even when the test binary links optimized engine objects.

A preprocessor check confirms the release endgame translation unit contains no `lookups` increment while the #625 baseline does.

## Correctness validation

- development `magpie_test tt` passes and observes the expected counters
- `no_pgo_release` `magpie_test tt` passes
- `no_pgo_release` `magpie_test endgame` passes
- exact values matched across every fixed-depth benchmark position and round
- deterministic one-thread searches had identical node counts

The counters are not search inputs, so release search behavior is unchanged. Diagnostic code that reads a TT populated by `NDEBUG` engine code will see zeroed statistics by design.

## Performance assessment

This did not produce the expected endgame speedup on the Mac mini.

All runs used `no_pgo_release`, CSW24 RIT on, 4-ply fixed-depth endgames, and five interleaved paired rounds.

| Threads / battery | Exact disagreements | Node behavior | Elapsed result |
|---|---:|---|---:|
| 10 threads / 20 positions | 0 | scheduling noise around flat | middle-three mean **-0.05%**; paired-position median **-0.85%** |
| 1 thread / 10 positions | 0 | exactly identical | middle-three mean **+0.14%** |

All results are below the approximately 1% resolution of this protocol. The PR therefore makes no performance claim despite removing real atomic instructions. It is a small release-path cleanup, not a demonstrated endgame win.

## Review focus

The main semantic question is whether any production consumer expects these diagnostic totals from a release-built engine. Source search found no production reads; current reads are tests and benchmark diagnostics. The table layout is unchanged, limiting the scope to update side effects.

### DIFF
--- a/src/ent/transposition_table.h
+++ b/src/ent/transposition_table.h
@@ -83,6 +83,18 @@ typedef struct TranspositionTable {
   atomic_int t2_collisions;
 } TranspositionTable;
 
+// TT statistics are useful in tests and diagnostic builds, but updating these
+// shared counters on every lookup and store serializes release search threads
+// on one cache line. Optimized test files are compiled without NDEBUG, so the
+// counter tests still exercise this path even when they link against release
+// engine objects.
+#ifdef NDEBUG
+#define TT_STAT_INCREMENT(counter) ((void)0)
+#else
+#define TT_STAT_INCREMENT(counter)                                             \
+  atomic_fetch_add_explicit(&(counter), 1, memory_order_relaxed)
+#endif
+
 static inline TranspositionTable *
 transposition_table_create(double fraction_of_memory) {
   TranspositionTable *tt = malloc_or_die(sizeof(TranspositionTable));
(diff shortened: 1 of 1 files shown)


olaugh:
Closing this: the measured performance was flat, so this release-only TT-stats change is not worth carrying.

PR #628, opened 2026-07-24
Disable RIT during leave generation

olaugh:
## Summary

- force both players’ RIT settings off before LeaveGen loads lexicon-dependent data
- unload a previously active RIT and ignore shared or per-player inline RIT enablement
- document why RIT is incompatible with LeaveGen
- add regression coverage for existing, shared, and per-player RIT settings

## Testing

- `make magpie_test`
- `./bin/magpie_test ap_rest`
- `python3 format.py config.c`
- `python3 format.py autoplay_test.c`


jvc56 (approved):

src/impl/config.c
```diff
@@ -5917,6 +5919,13 @@ void config_load_lexicon_dependent_data(
     }
   }
 
+  // Rack info tables cache KLV-derived leave values. Leave generation updates
+  // the KLV after every generation, so an RIT would become stale immediately.
+  if (disable_rit) {
```
olaugh, line 5924: One behavioral consequence worth a decision: this forces `use_when_available` off in the persistent `players_data`, and the test asserts it's still off after leavegen returns. So a session that had `-rit true`, runs a leavegen, then runs a sim will silently continue without RIT (~2.5–5% sim throughput) until the user re-specifies `-rit true`. Either restoring the pre-leavegen setting after generation completes, or a sentence in the help text saying the disable persists for the session, would keep that from surprising someone.

olaugh (commented):
This is really two fixes, and the quieter one matters more: the `klv_mark_leave_values_mutated` bump in `rack_list_write_to_klv` closes a real quality bug — before it, generations 2+ could generate moves against stale leave values cached in MoveGen's subrack and anchor-bound caches, silently degrading every leavegen run that reused a thread's MoveGen across generations. The exact-match gen-2/gen-3 CSV fixtures pin the full generation boundary, which is exactly the right regression shape for this. The RIT disable itself is correctly threaded (exec-token-gated at the one load site, covering inherited, shared-inline, and per-player enablement, with tests for all three). LGTM.

PR #629, opened 2026-07-25
Make sim/endgame/peg time limits independent; default endgame/peg to unlimited

jvc56:
## Summary
- Previously `-etlim`/`-pegtlim` fell back to `-tlim` when left at their 0 default, and `analyze` additionally bounded an unset `-etlim` with `-tlim`. This meant `-tlim`'s 60s default silently capped peg (and endgame, via analyze) solves too — the reported symptom was peg just stopping at 60s for no obvious reason.
- Removed the `-etlim`/`-pegtlim` → `-tlim` fallback entirely. `-tlim`, `-etlim`, and `-pegtlim` are now fully independent: `-tlim` only affects sim, `-etlim` only endgame, `-pegtlim` only peg.
- `-tlim` keeps its 60s default (sim is still capped by default). The plain `endgame`/`peg` commands now default `-etlim`/`-pegtlim` to 0 = genuinely unlimited (no fallback), so a standalone endgame/peg solve runs to completion unless the user sets a limit.
- `analyze` runs many turns per game, so an unbounded per-turn solve is a worse default there than for a single standalone solve. `analyze` now applies its own defaults when the user hasn't set `-etlim`/`-pegtlim`: **10 minutes** per endgame turn, **30 minutes** per peg turn. Setting `-etlim`/`-pegtlim` explicitly still overrides these for `analyze` as for any other command.
- Added a `status` field to `PegResult` (finished/interrupted/running), populated the same way `EndgameResults.status` is, and print it in peg's summary line(s) so a capped/interrupted solve is distinguishable from a completed one — matching endgame's existing status display. Both endgame's and peg's status now report "running" (not "none") while a solve is in progress, consistent with sim's existing "Running" status.

## Test plan
- [x] `make magpie BUILD=release` builds clean
- [x] `make magpie_test BUILD=release` builds clean
- [x] `./bin/magpie_test peg` passes, output shows `status: finished`/`running` in peg summary lines
- [x] `./bin/magpie_test endgame` passes, `status: running` shown for in-progress PVs, `status: interrupted` when time-limited
- [x] `./bin/magpie_test sim` passes
- [x] `./bin/magpie_test config` passes
- [x] `./bin/magpie_test analyze` passes
- [x] `./bin/magpie_test analyze_sim` passes (updated to set `-etlim 10 -pegtlim 20` explicitly so it doesn't inherit `analyze`'s new 600s/1800s defaults; runs in ~33s)
- [x] `./tidy.sh` shows no new warnings introduced beyond what's already present on `main`
- [x] `python3 find_circ_deps.py` reports no cycles
- [x] `python3 format.py --write` run (clang-format-20 unavailable locally; include ordering unaffected)


olaugh (changes requested):
Two issues need fixing before merge:

1. `src/impl/config.c` (`config_fill_analyze_args`): the fallback is selected from the numeric value alone (`*_time_limit_seconds == 0`). That cannot distinguish an option that was never set from an explicit `-etlim 0` / `-pegtlim 0`. Since zero is documented as unlimited, explicitly requesting zero under `analyze` unexpectedly becomes 600s/1800s, and contradicts the PR description that explicit settings override the analyze defaults. Please track whether each option was explicitly supplied (persistently, since settings can be set before the `analyze` command) and apply the analyze default only when it was not supplied. Add coverage for explicit zero as well as the unset case.

2. `src/impl/peg.c` (`peg_solve`, final status assignment): a solve that hits `deadline_ns` without a user interrupt falls through to `PEG_RESULT_STATUS_FINISHED`. The PR says the new status makes capped solves distinguishable from completed ones, but internal budget expiry is not reflected in the status; the output can therefore say `status: finished` after a partial/time-capped stage. Track deadline exhaustion and report it as interrupted (or introduce a dedicated timed-out status), with a test that uses `-pegtlim` and asserts the printed status.

All current GitHub checks are green, but neither edge case appears to be covered.

jvc56:
Addressed both issues in a74cfcfd:

1. **`config_fill_analyze_args` explicit-zero handling** — added `endgame_time_limit_explicitly_set`/`peg_time_limit_explicitly_set` bools to `Config`, set at parse time (persisted across commands, same as `peg_only_str`) whenever `-etlim`/`-pegtlim` is actually supplied. `config_fill_analyze_args` now branches on these flags instead of `== 0`, so an explicit `-etlim 0`/`-pegtlim 0` stays unlimited under `analyze` while a truly-unset option still gets the 600s/1800s default. Covered by `test_analyze_etlim_pegtlim_explicit_zero`.

2. **Deadline-capped solves reported as FINISHED** — rather than folding this into `INTERRUPTED`, added a dedicated `TIMEOUT` status to both `PegResult` and `EndgameResults` (distinct from `INTERRUPTED`, which is now reserved for genuine user interrupts). `peg_solve` tracks a `deadline_exceeded` flag across all three of its budget-cutoff sites; `endgame_solve`/`endgame_solve_inline` track the same on `EndgameCtx`, set at each of the three time-triggered early exits in `iterative_deepening` (per-depth deadline, soft-limit convergence, projected hard-limit overflow) — all of which mean the search didn't reach the full requested depth, so all three now report `timeout`. Covered by `test_peg_timeout_status` and `test_endgame_timeout_status`.

`make magpie`/`magpie_test` (BUILD=release) build clean, `./bin/magpie_test {peg,endgame,config,analyze,analyze_sim,sim}` all pass, `./tidy.sh` shows no new warnings vs. `main`, `find_circ_deps.py` reports no cycles. (Couldn't run `./cppcheck.sh` in this environment — no `cmake` available locally — worth a look in CI.)

PR #630, opened 2026-07-27
Add contextual KLV3 leave values

olaugh:
## Summary

Adds a draft KLV3 leave format and runtime: an ordinary KLV2 baseline plus a
small additive model conditioned on the public unseen pool and draw count.

For held leave `L`, public unseen multiset `U`, unseen count `N`, and draw
count `d`:

```text
V(L,U,N,d) = KLV2(L)
           + sum_h L[h] * (
                 bias[pool_bin(N),d,h]
               + sum_u weight[d,h,u] * U[u] / N)
```

The implementation:

- reads/writes the versioned `.klv3` trailer while preserving the KLV2 body;
- evaluates contextual placement and exchange leaves, shadow bounds, WMP
  leaves, and truncated-sim horizon residuals;
- includes a reproducible corpus recorder and offline trainer;
- preserves model-independent RIT word facts with a 407 MiB word-only RIT;
- fingerprints KLV-derived RIT data and invalidates fingerprints when a KLV is
  mutated;
- includes experimental p99 capped-RIT and hybrid KLV2-fallback paths;
- adds an on-demand oracle harness used for the policy experiments below.

Generated KLV/RIT files and experiment logs are intentionally not committed.
The format, training command, performance measurements, and artifact hashes
are documented in `notes/KLV3.md`.

### Static play

The selected interaction-only CSW24 model was trained from 50,000 static
games / 977,465 full-rack rows and tested in 250,000 paired static games:

- KLV3 wins: 249,823
- KLV2 wins: 248,234
- ties: 1,943
- score: **+0.4421 points/game** for KLV3
- roughly +1.1 Elo

A zero-coefficient KLV3 produced bit-identical games to KLV2.

### Runtime

In no-PGO, 10-thread simulation:

- exact sparse KLV3: 0.8296x KLV2 throughput;
- hybrid threshold 1.5: 0.9168x KLV2 throughput;
- word-only RIT versus full RIT for word facts: no measurable throughput
  difference; its benefit is footprint and startup I/O.

The p99 capped-RIT experiment was negative: it added about 3% overall cost and
is not enabled by default.

### Simulation policy

With exactly 1,500 samples per actual arm and a neutral nested oracle, KLV2
and corrected contextual-horizon KLV3 disagreed 500 times in 13,074 positions.
KLV3's conditional utility delta was -0.000420 (SEM 0.000284, p about 0.14).
There is no evidence that paying the throughput cost for contextual KLV3
throughout rollouts helps.

KLV3 was much more promising as a candidate selector for otherwise-KLV2 sims.
After removing the shared plays from independent top-15 lists and simming only
the unique fringe under KLV2:

- KLV3 candidate wins: 283-217 (exact p = 0.00361);
- conditional regret reduction: **0.002856 utility**
  (SEM 0.000723, p = 0.000089);
- full top-15 overlap: 849 / 1,349 source positions.

That is a marginal-candidate result, not an end-to-end strength estimate:
discarded shared candidates could beat both unique nominees. It justifies a
full-system KLV3-top-15/KLV2-rollout experiment.

## Validation

- `make -j10 BUILD=no_pgo_release magpie magpie_test`
- `./bin/magpie_test klv`
- `./bin/magpie_test rit`
- `./bin/magpie_test rv`
- `./bin/magpie_test cv`
- Python syntax checks for both new tools
- focused movegen, sim, autoplay, and config suites ran through the relevant
  CSW21 coverage but the local all-data tests later stop because this checkout
  does not contain NWL20/FRA20 artifacts
- zero-coefficient fixed-seed equality, including autoplay at 1 through 11
  threads

## Review notes

This is deliberately a draft and a large experimental PR. The production
format/runtime, training tooling, and oracle harness are kept together for
reproducibility now; the harness can be split before merge if that makes review
materially easier.

Stacked on #623.


src/impl/move_gen.c
```diff
@@ -538,6 +538,160 @@ static void record_best_exchange_from_table(MoveGen *gen) {
...
+  }
+  return count;
+#endif
+}
+
+// Contextualize the RIT's 128 KLV2 base leaves with a flat, fixed-size loop.
+// This is substantially cheaper than rebuilding a Rack and walking the KLV
+// trie for every canonical subset. It also finds the exact best contextual
+// exchange while the values are hot, so BEST movegen records it directly.
+static void fill_contextual_rit_leaves_and_record_exchanges(MoveGen *gen,
```
olaugh, line 575: Shadow-bound admissibility question: this computes contextual leave values and contextual `best_leaves`, and the shadow path's comment says exact-KLV2/structural-KLV3 bounds can't exceed the current context's maxima. Is there a guard preventing a *full* RIT (with leave-derived anchor upper bounds, computed pool-independently) from being combined with an active KLV3 model? If a user loads a legacy full RIT alongside a `.klv3`, those cached bounds could underestimate the contextual best and prune the true best move. If the word-only RIT is the only configuration KLV3 accepts, making that an explicit validation error (rather than a convention) would lock the invariant in.

olaugh (commented):
Reviewed at the architecture level given the draft status and scale (~7k lines); pushed a commit fixing the three cppcheck findings that were failing CI (unreachable-after-log_fatal array write in `klv3_get_rack_adjustment_range`, uninitialized `root_values` in the oracle test, side-effect lookup inside `assert` in the RIT test) and merged it forward into #631.

What stands out most is the experimental discipline: the zero-coefficient bit-identity check gives a hard guarantee that the machinery is inert when the model is, the 250k paired-game static result is properly powered for a +0.44 pts/game effect, and the sim-policy section reports its own null result (rollout-time contextual eval doesn't pay) alongside the significant one (candidate selection, p=0.000089 on conditional regret). Honest negatives recorded in the PR body — the p99 capped-RIT regression, the throughput costs — make the positive results credible. The word-only RIT is also the structurally right answer to a subtle staleness class: cached leave-derived bounds are pool-dependent under a contextual model, so keeping only model-independent word facts in the RIT (with KLV fingerprinting for the rest) removes that hazard rather than patching it.

Two requests before this leaves draft:

1. **Split it.** CLAUDE.md's own rule ("Large PRs with multiple concerns get closed") applies; the natural seams are clean: (a) `.klv3` format + runtime eval, (b) word-only RIT + KLV fingerprint invalidation, (c) corpus recorder + `train_klv3.py`, (d) the oracle harness test. The body already offers this — I'd take it up on it.
2. The headline next experiment (KLV3 as top-15 candidate selector over KLV2 rollouts) is well-motivated by the 283–217 fringe result; worth stating in the PR body whether that gates merging or lands separately.

PR #631, opened 2026-07-27
Forecast terminal spread from game state

olaugh:
## Summary

Stacked on #630.

MAGPIE currently has a horizon win-probability table, but no corresponding
forecast of terminal spread. In particular, total unseen tiles do not retain
the end-of-bag timing represented by the exact bag size and the two rack
sizes. This PR adds an experimental, opt-in terminal-spread model:

```
E[final spread | state] =
    current spread + mean swing[bag][on-turn rack][off-turn rack]
```

Each cell also stores expected remaining turns for each player and its sample
count. Those are diagnostics that let us validate the timing premise
directly. The standard English table is 89 KiB and lookup is O(1).

Use `-spreadforecast NAME` to load `data/strategy/NAME.sfc`, or
`-spreadforecast none` to disable it. The default remains off.

## Runtime semantics

At a nonterminal sim horizon, the forecast supplies:

- the residual recorded in equity/leftover statistics; and
- the spread input to blended sim utility.

It deliberately does **not** alter `-winpct`: win probability continues to use
the ordinary horizon spread plus leave residual. That isolates this experiment
and avoids silently invalidating the existing probability calibration.
Terminal samples continue to use actual final spread.

This creates the spread-scale API that a later static-eval sim prior can use,
but this PR does not add that prior.

## Training and format

- FJ recording now retains bag-empty positions.
- `tools/train_spread_forecast.py` splits whole games 80/20, shrinks sparse
  rack states toward their bag-count mean, reports held-out spread and turn
  errors by phase, and writes the compact versioned `.sfc`.
- KLV3 training explicitly ignores the new zero-draw post-bag rows, preserving
  its prior training semantics.
- The generated model and training corpus remain ignored data artifacts.

## Held-out forecast result

A fresh 50,000-game CSW24 static corpus produced 1,105,546 positions. Results
below are on the held-out 219,541 positions:

| Stratum | N | Forecast MAE | Current-spread MAE | Forecast RMSE | Current-spread RMSE |
| --- | ---: | ---: | ---: | ---: | ---: |
| All | 219,541 | **54.221** | 57.393 | **72.460** | 75.539 |
| Bag empty | 25,601 | **17.446** | 24.555 | **26.442** | 34.335 |
| Bag 1–7 | 17,051 | **33.640** | 37.637 | **44.164** | 47.990 |
| Bag 8–20 | 30,301 | **42.371** | 45.734 | **53.751** | 57.686 |
| Bag 21+ | 146,588 | **65.487** | 67.835 | **83.170** | 85.932 |

Overall MAE improves 5.5%. Bag-empty MAE improves 29.0%, the expected
signature if separate rack sizes capture end-of-game timing. Bag-empty
expected-turn MAE is 0.294 turns for the player on turn and 0.354 for the
opponent.

This is predictive validation, not a playing-strength result. The model still
omits rack composition, board openness, position-specific scoring rate, and
score-dependent strategy.

## Runtime

Nine interleaved fixed-trajectory simbench pairs, no-PGO release, 10 threads,
RIT on, 0.2-second turn limits:

- median forecast-on/off throughput ratio: 0.9989
- mean of the middle three ratios: 0.9978 (-0.22%)

That is below the benchmark's roughly 1% resolution; no runtime cost is
currently measurable.

## Validation

- dev build with ASan/UBSan and `-Werror`
- GCC 15 compilation with the CI warning set
- `spreadforecast` loader/lookup/config tests
- `rv` tests
- repository formatting check
- Python syntax checks

The local full `sim`/`autoplay` suites proceed until tests requiring the
missing local NWL20/FRA20 artifacts. CI supplies those files.

## Next experiment

Run equal-position, equal-sample sims with the forecast off/on, and oracle only
candidate-ranking disagreements. Add static evaluation as a prior only if this
forecast improves those decisions; the prior's calibration should be a
separate experiment.


src/impl/random_variable.c
```diff
@@ -535,8 +538,23 @@ double rv_sim_sample(RandomVariables *rvs, const uint64_t play_index,
...
+  Equity projected_residual = leftover;
+  if (simmer->spread_forecast != NULL &&
+      game_get_game_end_reason(game) == GAME_END_REASON_NONE) {
+    projected_spread = spread_forecast_get(simmer->spread_forecast, game,
+                                           simmer->initial_player);
+    const int64_t residual = (int64_t)projected_spread - (int64_t)spread;
+    if (residual > EQUITY_MAX_VALUE || residual < EQUITY_MIN_VALUE) {
+      log_fatal("spread forecast residual out of Equity range");
+    }
+    projected_residual = (Equity)residual;
```
olaugh, line 551: The forecast residual *replaces* the leave-value `leftover` in the equity sample (it becomes `projected_spread − initial_spread`). Since the forecast cell is conditioned only on bag/rack counts, this discards the position-specific leave signal that `leftover` carried — two horizons with identical counts but a QI-vs-SATIRE on-turn rack get the same projected swing. Was that deliberate (avoiding double-counting, since the table's mean swing already includes average leave effects), or would centering help — i.e., forecast + (leftover − mean leave contribution)? If deliberate, a comment here saying the replacement is intentional and why would preempt the obvious "shouldn't these add?" question. The held-out MAE win suggests the trade is fine on average, but the sim cares about ranking arms within one position, where the discarded leave term is exactly the discriminating signal.

olaugh (commented):
Clean, well-isolated experiment. Things I verified: the `.sfc` loader validates dimensions against the header, checks every cell for finiteness and non-negative turn counts, detects trailing data, and `double_to_equity` log_fatals on out-of-range — so a malformed file fails loudly rather than corrupting equity. POV sign handling in `spread_forecast_get` is correct (cell is on-turn-relative, negated for the other perspective), terminal samples bypass the forecast, and both `rv_sim_create` and `rv_sim_reset` carry the pointer. Keeping `-winpct` on the raw horizon spread is the right isolation call — it means a bad forecast can only hurt the spread half of blended utility, never the probability calibration. The bag-empty stratum result (MAE 17.4 vs 24.6) is the cleanest direct validation of the end-of-bag-timing premise the PR is built on.

One modeling question inline; otherwise this looks ready to graduate from draft once #630 settles underneath it. Note I merged #630's cppcheck fixes forward into this branch to get CI green.

PR #636, opened 2026-07-30
Use first-win searches for direct PEG endgame leaves

jvc56:
## What this does

Uses the existing narrow endgame window for direct bag-empty PEG leaves. Nested leaves retain full-spread searches because their values are averaged before classification. The PEG CLI enables this by default; `-pegfw false` restores full-spread leaves throughout PEG; `-efw` separately exposes the endgame option.

Narrow-window values can be bounds. They can change spread estimates, move rankings, and later-stage survivors. The payoff is therefore judged by the quality of the chosen move under an equal time budget.

## Performance

Independent holdouts compare main and the refined candidate with a common **main-branch oracle at five times the arm budget, with first-win disabled**. Utility is `U = win_pct + 0.0001 * mean_spread`. Negative candidate-minus-main regret favors this PR.

| Holdout | Positions | Main mean regret | Candidate mean regret | Paired difference | Position-bootstrap 95% interval |
| --- | ---: | ---: | ---: | ---: | --- |
| 2-second arms / 10-second oracle | 128 | 0.027293 | 0.025848 | -0.001445 | [-0.002942, -0.000273] |
| 5-second arms / 25-second oracle | 64 | 0.013018 | 0.008290 | -0.004728 | [-0.011355, +0.000007] |

The primary result favors the PR with an interval excluding zero. The separately prescribed longer-budget result has a favorable mean but an interval narrowly crossing zero, so it is inconclusive on utility by itself. Median paired difference is zero in both holdouts: most moves agree.

- **2 seconds:** 11 lower-regret choices, 106 ties, 11 higher-regret choices. Win-probability regret improves on 6 positions and ties on 122; none worsens. Mean win-regret reduction is **0.1497 percentage points**, 95% interval [0.0326, 0.2984].
- **5 seconds:** 6 lower-regret choices, 52 ties, 6 higher-regret choices. Win-probability regret improves on 4 positions and ties on 60; none worsens. Mean win-regret reduction is **0.4737 percentage points**, 95% interval [0.0071, 1.1318].

All utility losses in these samples are spread tie-break differences. Results depend on position and budget: the one-tile stratum has a small mean utility loss at two seconds, while the other shorter-budget strata improve or tie. These are bounded-oracle estimates, not game-theoretic values or a tournament win-rate measurement.

### Measurement details

M4 Mac mini, 10 cores, 16 GiB, macOS 15.7.4, Apple Clang 17. Independent production static-autoplay PGO training (16 games, four training threads), identical CSW24 assets, WMP/RIT/WIT enabled, eight solve threads, fresh process/solver TTs, total TT allocation below half RAM. A persistent `bird` sync process used roughly one core; per-position host snapshots are retained. No build/test/profile jobs competed with timed comparisons.

Main `64019d5d` and candidate `d3710c35` are the frozen measured source identities. Main `0be09d0a` differs from that baseline only in AGENTS.md. Subsequent edits correct benchmark argument parsing and comments; the production search behavior measured here is unchanged.

The primary corpus has 32 fresh static-self-play positions per bag size 1–4; the secondary has 16 per bag size. Bag strata are interleaved and arm order alternates. The nested-full-spread refinement was selected on a separate 16-position discovery set before either holdout. Intervals use 10,000 position-paired bootstrap resamples with a fixed seed; the holdouts are reported separately.

Both selected moves are protected in a top-32, three-stage common oracle, including when the arms agree. Score at the deepest stage where both values completed within the oracle deadline. **192/192 positions have coverage; all 384 selected arm moves completed before their budgets.** No positions were replaced or excluded. Common stages 0/1/2/3 were 55/56/16/1 in the primary and 21/33/5/5 in the secondary, so much of the oracle evidence remains shallow.

Cancellation and cleanup can finish after the search deadline. Median main/candidate/oracle return times were 2.201/2.131/10.293 seconds and 5.205/5.151/25.262 seconds; only timestamped values available inside the search budgets entered the comparison.

## Profiling

Four paired observer workloads (first secondary position from each bag stratum), alternating order, six-second samples starting one second into a 12-second solve. Same frozen PGO binaries as the holdouts. Addresses were symbolized against those exact binaries; system waits are excluded from engine self-sample shares.

The changed path is `peg_eval_leaf` → `endgame_solve_inline` → `abdada_negamax`; `peg_nested_endgame_value` continues using full-spread searches. Most CPU samples lie in move generation beneath these searches.

| Engine function group | Main self-sample share | Candidate share |
| --- | ---: | ---: |
| `recursive_gen_small` + `go_on_small` | 26.84% | 21.87% |
| `shadow_play_for_anchor_small` + `shadow_play_right_small` | 13.40% | 16.41% |
| `shadow_play_for_anchor` | 14.18% | 14.18% |
| `generate_moves` | 13.02% | 13.85% |
| `game_gen_cross_set` | 4.82% | 5.59% |

The mechanism changes which descendants are searched and how far the cascade progresses within its budget. These are workload shares, **not per-function speedups**. One observer capture hit disk exhaustion and produced an empty file; it was retained as failed and replaced by a verified capture, without changing any timing or regret results.

## Correctness

- PEG, endgame, and config suites pass on the production-PGO refinement.
- With first-win disabled and one solve thread, all **469 candidate/stage values match main exactly**. The initial eight-thread control and the final-parser check also matched all 469 values.
- Three repeated eight-thread off-path timing pairs center near zero (**+0.24%** geometric-mean candidate speedup, individual pairs −2.51% to +1.97%). This is a small one-position control, not a broad throughput claim.
- **Review caveat:** repeated eight-thread candidate runs varied by 0.25 spread points in two non-selected candidate/stage estimates. The best move and all win estimates stayed unchanged. Repeated main runs did not reproduce that variation, so its cause has not been isolated and this PR does not claim parallel bit-for-bit invariance. All logs are retained; the diagnostic runs are excluded from the holdout results.
- The C `pegregretbench` driver synchronizes concurrent callbacks with the project's mutex helpers and records availability timestamps for budget auditing.
- Full cppcheck and dependency-cycle checks pass. The benchmark uses the project's checked integer parser.
- First-win bounds are not presented as exact spread values; nested averaging retains full-spread searches.

## Reproduction

Build with `make release NPROCS=2 PGO_TRAIN_THREADS=4 PGO_TRAIN_GAMES=16 PGO_CC=clang`, then `make magpie_test BUILD=pgo_use CC=clang NPROCS=2`.

Run `pegregretbench` with `PEGP_CGP`, `PEGP_SECONDS`, `PEGP_THREADS=8`, and `PEGP_FIRSTWIN`. For the common oracle set `PEGP_ORACLE=1`, `PEGP_FIRSTWIN=0`, and `PEGP_PROTECT` to the comma-separated chosen UCGI moves, with five times the arm budget. Generate primary corpora using `PEGP_GENERATE=1`, `PEGP_FILE`, `PEGP_BAG`, `PEGP_COUNT=32`, and seeds `630200101 + 1000 * bag`; secondary uses count 16 and seeds `630300101 + 1000 * bag`.

Full protocol, source bundle, frozen binaries/PGO profiles, CGPs, callback traces, host snapshots, bootstrap summaries, control diagnostics, and raw profiles are preserved in `/Users/olaugh/sources/sep08-peg-firstwin-evidence`. Production benchmark/data generation is C; the evidence scripts only orchestrate and summarize runs.


jvc56:
`pegfw` should probably default to false since using true performs much worse when used with pruning stages.

PR #650, opened 2026-08-12
Fix two analyze/endgame bugs found via real tournament data

truthmane:
## Summary
- Fix a row/col double-swap in `augment_single_tile_actual_move` (`src/impl/endgame.c`): when the mover has exactly one tile left and plays it vertically, the actual move was silently re-encoded at the wrong board square before being appended as a root candidate, so the endgame solver could never match it — surfacing as `endgame solver failed to find the actual move among root moves` (or, without that check, evaluating the wrong move entirely). `Move.row_start`/`col_start` are stored pre-swapped for vertical plays; `small_move_set_all` performs that same swap internally, so passing the already-swapped fields through double-swapped them. Fixed by undoing the swap before calling `small_move_set_all`.
- Fix `analyze`'s inference-gating logic (`src/impl/analyze.c`) to also skip inference when the opponent's most recent move was a pass, not just when they played a full rack. The guard was only checking the full-rack-play case ("empty leave, nothing to infer"), so a plain pass (no tiles played or exchanged) fell through and hit `inference_worker`'s own hard rejection of that combination, crashing the whole `analyze` run.

Both bugs were found by running `analyze` over a real 30-game tournament directory and hitting a hard `log_fatal`/error partway through. Each is root-caused, fixed, and reproduced/re-verified against the actual failing game.

## Test plan
- [x] `make magpie_test BUILD=release && ./bin/magpie_test` — full suite passes
- [x] `./bin/magpie_test endgame analyze` — passes on this branch specifically
- [x] `python3 format.py --write` — no changes needed
- [x] `./cppcheck.sh` — clean
- [x] Reproduced the endgame crash on a real GCG with a vertical one-tile final play; confirmed fixed after the patch (the previously-unmatched move now correctly evaluates as best, WPL 0.00)
- [x] Reproduced the inference crash on a real GCG containing an opponent pass; confirmed fixed after the patch
- [x] Reran the full 30-game tournament directory end-to-end with both fixes applied: 30/30 games analyze successfully (previously failed at game 7 and game 30)

### DIFF
--- a/src/impl/analyze.c
+++ b/src/impl/analyze.c
@@ -940,13 +940,14 @@ static void analyze_with_sim(const GameEvent *event, TurnResult *turn_result,
       player_get_rack(game_get_player(ctx->game, 1 - player_on_turn_index));
 
   // Suppress inference when the opponent's rack is already fully known, or
-  // when the opponent just played a full rack (their leave was empty so there
-  // is nothing to infer from).
+  // when the opponent's most recent move left nothing to infer from: a full
+  // rack played (empty leave) or a pass (no tiles played or exchanged --
+  // inference_worker itself rejects that combination, see inference.c).
   const bool opp_rack_fully_known =
       rack_get_total_letters(known_opp_rack) == RACK_SIZE;
 
   const int opponent_index = 1 - player_on_turn_index;
-  bool opp_played_full_rack = false;
+  bool opp_left_nothing_to_infer = false;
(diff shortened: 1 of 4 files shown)


jvc56:
Let's fix inference pass at inference.c instead of analyze since I think this is also a bug for annotation mode. When the inference logic gets a pass as an input it should just return an empty infer result without an error.

jvc56 (approved):
LGTM! You can merge this in.

PR #652, opened 2026-08-18
Let commit-by-index reuse a pending redo event's buffered results

truthmane:
## Summary

- After committing a move and then running "prev" to undo it, `com <index>` (and `cnote`/`cn`, which shares the same code path) could not commit a *different* move from the same generated/simmed candidates — it failed with "cannot commit move with index '2' with no generated moves", even though the analysis that produced the undone move was still sitting in the game history.
- Adds `config_get_redo_game_event`, which finds that pending "redo" event ("prev"/"goto" only move the `num_played_events` cursor back; they don't delete the event, which stays in place until a new commit overwrites it via `game_history_truncate_to_played_events`).
- Adds a commit-specific resolver, `config_resolve_current_results_for_commit`, used only by `parse_commit`'s commit-by-index branch. It deliberately does *not* fall back to the last-played event the way `shmoves`'s existing fallback does — that buffer's moves were already played onto squares the board no longer has room for, so reusing it for a second commit would just fail move validation instead of committing anything (this distinction is now covered by two separate resolver functions with their own comments explaining why).

## Test plan

- [x] `./bin/magpie_test config` passes, including the existing `test_config_game_event_results_follow_navigation` and `test_config_anno` cases that pin down the (unchanged) `shmoves` fallback behavior at position 0
- [x] Full `./bin/magpie_test` suite passes
- [x] Manually reproduced the reported scenario end-to-end (`gen` → `sim` → `com 1` → `prev` → `com 2`) and confirmed it now commits a different move with no error, for both `com <index>` and `cnote <index> ...` (`cn`)
- [x] `python3 format.py --write`, `./cppcheck.sh`, `./tidy.sh` all clean (one pre-existing `misc-include-cleaner` warning on an unrelated line, confirmed present on `main` too and unaffected by this change)

### DIFF
--- a/src/impl/config.c
+++ b/src/impl/config.c
@@ -3992,7 +3992,120 @@ static const GameEvent *config_get_current_game_event(const Config *config) {
   return game_history_get_event(config->game_history, num_played_events - 1);
 }
 
-char *impl_show_moves_or_sim_results(Config *config, ErrorStack *error_stack) {
+// The GameEvent that "prev"/"goto" most recently backed away from, if any.
+// "prev" and "goto" only move the num_played_events cursor; they don't
+// delete anything, so the event describing the move that was undone (the
+// one that used to be "current") is still physically present at
+// events[num_played_events] until either a new move is committed from here
+// (game_history_truncate_to_played_events discards it then) or "next"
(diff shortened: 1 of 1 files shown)


jvc56:
Fixed in #653

PR #654, opened 2026-08-26
Peg allow guaranteed bag empty

jvc56:


src/impl/peg.c
```diff
@@ -2421,23 +2450,41 @@ void peg_solve(const PegArgs *args, PegResult *out, ErrorStack *error_stack) {
...
+    // returns what it drew), and no tile placement can play more than
+    // RACK_SIZE tiles, so the guarantee is only reachable for bag_size in
+    // (PEG_MAX_BAG, RACK_SIZE].
+    //
+    // This can only ever be satisfied via a caller-supplied only_moves set
+    // (e.g. the pegonly "empty" positional value): the full root move list
+    // always includes a pass (0 tiles played, see gen_record_pass), which
+    // never empties the bag, so without only_moves the guarantee can never
+    // hold and there is nothing to check.
+    bool bag_emptying_guaranteed =
```
olaugh, line 2474: The carve-out branch is entered for both out-of-range sides, but the guarantee condition only bounds the bag from above. When `bag_size <= 0` (a true endgame, or an inconsistent CGP where known opponent tiles exceed the raw bag), `peg_move_empties_bag(move, bag_size)` is trivially true for any tile placement, so the position passes validation. Downstream, `k_drawn` goes nonpositive, `peg_enum_splits` enumerates zero scenarios, and the solve "succeeds" with win_pct 0.0 for every candidate — where main returned `ERROR_STATUS_PEG_BAG_OUT_OF_RANGE`.

Suggest:
```c
bool bag_emptying_guaranteed = bag_size > PEG_MAX_BAG && args->n_only_moves > 0;
```
That matches the comment's own "(PEG_MAX_BAG, RACK_SIZE]" claim, and the `<= RACK_SIZE` term becomes unnecessary since no placement can play more than RACK_SIZE tiles.

src/impl/config.c
```diff
@@ -3483,7 +3486,11 @@ config_generate_peg_bag_emptying_moves(const Config *config,
   };
   generate_moves(&gen_args);
 
-  const int bag_size = bag_get_letters(game_get_bag(config->game));
+  // The PEG-effective bag size (matching peg_solve's own computation), not
+  // the raw bag_get_letters count: when the opponent's rack isn't fully
+  // known, those tiles sit in the raw bag too, but they are not part of the
+  // real bag peg_solve will enumerate.
+  const int bag_size = peg_compute_bag_size(config->game);
```
olaugh, line 3495: The filter a few lines below keeps moves by `tiles_played >= bag_size` alone, without the `GAME_EVENT_TILE_PLACEMENT_MOVE` check that `peg_move_empties_bag` has. Movegen's exchange gate (raw bag + opp rack >= 2×RACK_SIZE) is exactly "effective bag >= 7", so at effective bag 7 the full-rack exchange is generated, kept by this filter, and then fails the guarantee loop in `peg_solve` — rejecting the whole solve even when genuine bag-emptying placements exist. So `peg empty` can never work at the top of the range this PR enables. The CLI test only covers the 5-bag board, where no exchange is generated, so this path is untested.

Suggest exporting `peg_move_empties_bag` in peg.h (next to the `peg_compute_bag_size` declaration this PR already adds) and calling it here, so the predicate has one definition — plus a `peg empty` CLI test on the 7-bag board.

src/impl/config.c
```diff
@@ -3445,8 +3445,11 @@ static ValidatedMoves *config_parse_peg_move_list(const Config *config,
     return NULL;
   }
   // Exchanges need >= 7 tiles in the bag, so validated_moves_create above
-  // already rejects them in any PEG position (bag <= PEG_MAX_BAG); tile plays
-  // and the pass parse through.
+  // rejects them outright in the normal PEG range (bag <= PEG_MAX_BAG); tile
```
olaugh: Small correction to the new comment: exchange validation uses the *raw* bag (`is_exchange_allowed`: `bag_get_letters >= RACK_SIZE`), so with an unknown opponent rack (raw = effective + 7) an exchange validates at any effective bag size, not just at RACK_SIZE — and in the normal 1..4 range the guarantee check never runs, so `peg <exchange>` flows into the candidate set unchecked. That feed-through predates this PR, but since the comment is being rewritten anyway, it might be worth adding the same move-type check to parsed only_moves rather than restating the old claim.

olaugh (commented):
A few findings from a close read of the bag-emptying-guarantee change. The core idea looks sound (the array-capacity math checks out, and the no-only_moves early rejection is right since the full move list always contains a pass), but I think there are two real bugs in the new logic — inline comments below. Two more notes that don't anchor to changed lines:

**peg.h contract not updated:** The PegArgs docs ("Must have bag size in [PEG_MIN_BAG, PEG_MAX_BAG]", peg.h:103) and the module header ("larger positions are rejected by the solver", peg.h:17) now contradict the new exception — as written they'd justify someone restoring the strict check later, and nothing in the struct docs states the only_moves precondition for the relaxed range.

**Nested-peg arrays:** Leaving PegNestScenarioJob's mover_drawn/bag_remaining (and peg_nested_cand_value's locals) at PEG_MAX_BAG + 1 looks correct — in relaxed mode every scenario empties the bag, so the inner-peg path is unreachable — but that invariant is only enforced at solve entry. A one-line `assert(k_drawn <= PEG_MAX_BAG)` at the nested entry (or a comment explaining the exemption from the PEG_SCENARIO_ARRAY_CAP resize) would turn a future violation into a crash at the site instead of silent out-of-bounds writes.

PR #656, opened 2026-08-27
Add a `contribute` command: MAGPIE as a birdtest worker client

jvc56:
Makes MAGPIE able to contribute to [birdtest](https://github.com/jvc56/birdtest) on its own, so a contributor needs MAGPIE and nothing else — no Python, no Docker. Implements [MAGPIE-CLIENT.md](https://github.com/jvc56/birdtest/blob/implement-v1/MAGPIE-CLIENT.md).

```
magpie> contribute
```

## Why in-process

The existing client is a Python script that shells out to MAGPIE and parses its stdout. Every integration bug found so far has lived on that process boundary: invented flags, output that turned out to be aggregate-only, a rack-equity CSV written under a name the client didn't predict, and MAGPIE resolving its board layout from `./data` before parsing `-path`. Running the client in-process deletes that class of problem — task execution calls MAGPIE's own command API and reads results out of the result structs.

## Platform code stays in `src/compat`

That invariant currently holds exactly — nothing outside `src/compat` tests `_WIN32`, `__APPLE__` or `__wasm__` — and this is the largest new source of platform behaviour MAGPIE has taken on, so it was worth preserving deliberately. Four new compat pieces, each exposing a neutral API:

| File | Provides |
|---|---|
| `chttp.{h,c}` | HTTP + TLS: libcurl on POSIX, WinHTTP on Windows, stub on wasm |
| `crandom.h` | Cryptographic random bytes for the worker UUID |
| `cfile.h` | Owner-only permissions for the file holding the API key |
| `csleep.h` | Blocking sleep for poll and rate-limit backoff |

**libcurl is `dlopen`ed, not linked.** Linux ships no OS-level HTTP API and libcurl is what the platform provides; macOS ships it too, so one implementation covers both. Resolving it at runtime means a machine without libcurl still runs every offline command, and only `contribute` reports the problem — with the package name. Certificate verification is on and has no override.

Everything above compat is portable: vendored cJSON with an `ErrorStack`-aware wrapper, an HTTP client holding the retry policy (429 honours `Retry-After`; 5xx and transport failures back off exponentially), and the task loop.

## Settings come from a file

`contribute.txt` in the working directory, one `key value` per line. Never the command line: an API key there ends up in shell history and `ps` output, and contribution settings have no business mixed into `settings.txt`. The file is user-authored and never rewritten, except that a missing `uuid` is generated and **appended** as one line so comments and formatting survive. An unknown key is an error rather than a silent ignore, so a typo'd `apikey` can't quietly downgrade someone to anonymous.

## Also in here

- **`get_gcg.c` no longer shells out to the `curl` binary.** Its three call sites built shell command strings partly from caller-supplied identifiers and required curl to be installed. They now use the compat layer — which is also its first real exercise, verified by loading a live cross-tables game.
- **`autoplay_results_get_game_summary()`** exposes the game counts and score moments that were previously only reachable by parsing formatted output.
- **`-ldl` on Linux only**, since glibc before 2.34 keeps `dlopen` in a separate library.

## Testing

`magpie_test contribute` covers version comparison (including that `1.10 > 1.9`, which naive string comparison gets backwards), the JSON wrapper including a full `uint64` round trip, JSON escaping round-tripping through the parser, and settings parsing including the unknown-key and missing-server failures. `magpie_test autoplay` still passes.

Verified end to end against a local birdtest instance: claimed and completed real game-pair tasks, with the server recording results and crediting the worker. A job requiring MAGPIE 9.9.9 stops with `this job requires MAGPIE 9.9.9 but this build is 0.0.0`.

## Not done

- **`leave_generation` is not implemented** — it needs artifact download and reading the rack-equity table out of `RackList`. It reports that clearly and stops rather than failing obscurely.
- **`opening_rack_analysis` is written but unverified end to end.** Testing it needs a job whose lexicon MAGPIE has, and a full English rack enumeration is millions of tasks. Its error path is verified.
- **The WinHTTP backend is written but has not been compiled or run on Windows.**
- The async GUI status surface (state/progress/totals) is not implemented.

One design note worth flagging: a task that *fails* doesn't count toward `maxtasks`, so an unrunnable job would spin forever claiming and abandoning. There's now a consecutive-failure guard that stops after 5 with the last error.


jvc56:
Addressed review feedback in two follow-up commits:

**Design changes:**
- Removed `src/compat/cfile.h` — the contribution settings file needs no special permission handling; it's a plain file the contributor already controls, like `settings.txt` or a CGP file.
- Removed `src/compat/csleep.h` — all sleep/backoff calls now use the existing `ctime_nap()` from `src/compat/ctime.h`.
- Removed `src/compat/crandom.h` — the worker UUID is now **server-issued**, not client-generated. A request with no API key and no stored UUID sends no identity header at all; the server mints a UUID and returns it in the first successful task claim, and the client persists it from there. Implemented end-to-end, including the matching [birdtest backend change](https://github.com/jvc56/birdtest) (`WorkerIdentity` now mints an anonymous UUID server-side instead of requiring the client to supply one).
- Moved `src/util/cjson.{c,h}` → `src/compat/cjson.{c,h}` — the vendored cJSON source is itself platform-conditional (`#ifdef _WIN32`), so it belongs with the other platform-specific code. Excluded from cppcheck/tidy the same way `linenoise.c` (the other vendored file) already is.
- Fixed the `contribute` ↔ `config` circular dependency (`lint-fast`'s check): `contribute.h` no longer includes `config.h`. Following the pattern every other multi-step command already uses (`impl_autoplay`, `impl_peg`, `impl_leave_gen`, ...), the `Config`-aware orchestration now lives in `config.c` itself, calling into `contribute.c`'s `contribute_run()` through a small `ContributeConfigApi` vtable instead of passing `Config` across the module boundary.

**CI:** all 17 checks now pass, including `clang-tidy` — my first pass at that had used the machine's default `clang-tidy` (LLVM 10) rather than the CI-pinned `clang-tidy-18`, which missed several checks (`misc-include-cleaner`, `bugprone-casting-through-void`, `readability-avoid-nested-conditional-operator`, `cert-err33-c`) now fixed and verified against `clang-tidy-18` directly.

jvc56:
Reworked the `contribute` ↔ `config` architecture per review: removed `ContributeConfigApi` entirely and redistributed responsibilities so neither module needs the other's type, following the pattern every other multi-step command already uses (`impl_autoplay`/`config_autoplay`, `impl_load_gcg`/`get_gcg`, ...):

- **`contribute.c`** now owns only the birdtest HTTP/JSON protocol — claiming, heartbeating, submitting, the anonymous-UUID handshake — through a new `ContributeState` + `contribute_claim_task()`/`contribute_submit_result()`. It no longer includes anything Config-shaped.
- **`config.c`'s `impl_contribute()`** owns the loop and, between a claim and its matching submit, executes the task with direct methods — `config_autoplay()`, `game_load_cgp()`, `impl_move_gen()`, `impl_sim()`, `config_load_sort_type()`/`config_load_record_type()`, `config_load_lexicon_dependent_data()`, `convert()` for on-demand wordmap provisioning — never a command string.

Also fixed the small cleanups from the first round of review (moved `trim()` into `string_util`, `contribute_compare_versions()` no longer calls `strtol()` directly, `heartbeat_worker()` reuses its `StringBuilder`/`ErrorStack` across iterations instead of recreating them).

Verified the refactor against a local mock birdtest server (all 4 job types, the unknown-job-type/MAGPIE-too-old fatal-stop paths, and the anonymous-UUID handshake) under both release and ASAN/UBSAN builds. That surfaced two real, pre-existing bugs (not introduced by this refactor) which are now fixed:
1. A leak — the JSON-result `StringBuilder` was dumped but never destroyed.
2. A crash in `opening_rack_analysis`'s simulation path — it called `impl_sim()` without a move list ever having been generated first, which is presumably why the original PR flagged that path as "unverified end to end."

All 17 CI checks pass.

PR #657, opened 2026-08-30
endgame: solve every -etopk N line exactly, not just the best one

jvc56:
Every line shown by `-etopk N` is now a solved variation. Previously only the top line was searched to depth; the rest were one exact move plus a TT probe plus a greedy playout, which is why every line but the first showed the `|` separator right after its first move.

## What changed

After the workers join, each displayed root move is re-searched to full depth with a full window as a PV node, and its exact continuation is recorded. Because the TT is warm from the main search, only the PV chain is re-expanded — every off-PV node cuts off on a TT hit.

Resolution continues past K while the next candidate's upper-bound estimate still beats the worst exact value on the board, so the *set* of N lines is certain, not just each line's continuation. (A root move outside the multi-PV top-K only ever gets a narrowed window, so its `estimated_value` is a fail-low upper bound, and an inflated bound can promote a move that does not belong.)

Under a time limit the pass gets only the unspent remainder of the budget (PEG's `external_deadline_ns` still applies); on a user interrupt or an exhausted budget it stops and the previous TT+greedy fallback fills the remaining slots. `first_win_optim` and `initial_window_optim` solves, whose root values are deliberately bounds, skip it.

Drive-by: the root-move de-duplication read `solver->principal_variation`, which only `endgame_solve_inline` ever sets, so on the threaded path it compared against a stale line. It now reads `multi_pvs[0]`.

## Evidence

Release builds run as separate processes so each solve gets a cold, private TT (`-ttfraction 0.02`), no time limit, idle host. Timing uses `-threads 4`; output comparisons use `-threads 1`, because the parallel search is nondeterministic among equal-valued ties — repeated `-threads 4` runs of the *same* binary can report different moves, so identity is only meaningful single-threaded.

Positions (blank-free racks, all well under two minutes), taken from the existing endgame tests:

- **A** — `AEEIRUW/V`, CSW21 (`test_solve_standard`)
- **B** — `BGIV/DEHILOR`, NWL20 (`test_small_arena_realloc`)
- **D** — `BFLNOOR/EOSSSS`, CSW24 (`test_endgame_outplay_zobrist_overflow`)

### `-etopk 1` is untouched

`extract_multi_pvs` is only reached when `num_top_moves > 1`. Verified at `-threads 1`, with a base-vs-base control to show the comparison can detect a difference at all:

| cell | base vs base | base vs new |
| --- | --- | --- |
| A, 5 plies | same | **identical** |
| B, 5 plies | same | **identical** |
| D, 5 plies | same | **identical** |
| D, 7 plies | same | **identical** |

### Timing by `-etopk N`

`-threads 4`, mean of 2 reps. Sub-second cells report an absolute delta.

| position | plies | etopk | base (s) | new (s) | delta |
| --- | --- | --- | --- | --- | --- |
| A | 5 | 1 | 11.82 | 11.65 | -1.4% |
| A | 5 | 2 | 12.14 | 12.21 | +0.6% |
| A | 5 | 4 | 12.37 | 12.50 | +1.1% |
| A | 5 | 8 | 12.38 | 12.45 | +0.6% |
| A | 5 | 16 | 12.92 | 12.77 | -1.2% |
| D | 5 | 1 | 0.12 | 0.12 | -0.00s |
| D | 5 | 2 | 0.21 | 0.22 | +0.01s |
| D | 5 | 4 | 0.64 | 0.65 | +0.02s |
| D | 5 | 8 | 1.01 | 0.96 | -5.4% |
| D | 5 | 16 | 2.08 | 2.18 | +4.7% |
| D | 7 | 1 | 0.15 | 0.15 | +0.00s |
| D | 7 | 2 | 0.24 | 0.25 | +0.01s |
| D | 7 | 4 | 2.52 | 2.56 | +1.4% |
| D | 7 | 8 | 3.82 | 3.86 | +1.1% |
| D | 7 | 16 | 7.97 | 8.14 | +2.1% |
| A | 6 | 1 | 24.19 | 24.69 | +2.1% |
| A | 6 | 2 | 24.76 | 24.68 | -0.3% |
| A | 6 | 4 | 24.82 | 24.50 | -1.3% |
| A | 6 | 8 | 24.96 | 25.12 | +0.7% |
| A | 6 | 16 | 25.22 | 24.99 | -0.9% |

Noise on this host is ~±3%, so every cell is noise.

### Moves and values are unchanged

At `-etopk 8`, `-threads 1`, base and new agree exactly on both the set of lines and every value, on A/5, A/6, D/5 and D/7. What changes is that the continuations are solved instead of guessed. `-eplies 5 -etopk 6 -threads 1` on position D, before — only the top line is exact, and every other line is one exact move followed by a greedy playout, marked by `|`:

```
Play            V   S  Turn 2                   Turn 3      Turn 4          Turn 5      Turn 6
1C FLOR 19      -17 69 L8 SOSSES 26 (BNO +10)
9K ROLF 19      -20 66 | O7 SOSSES 29 (BNO +10)
15A LO 33       -22 64 | 9I (Z)OS 25            7K (V)OR 7  5C (KITH)ES 17  F7 F(U)B 8  L6 S(O)S 26 (N +2)
F7 B(U)FO 11    -23 63 | L8 SOSSES 26 (LNOR +8)
N4 LOOF 11      -25 61 | L8 SOSSES 26 (BNR +10)
11H (E)NROL 10  -26 60 | M9 SOSSES 20 (BFO +16)
```

After — every line is exact to the search depth. `|` now marks the 5-ply search boundary, and the lines that end the game before it carry no separator at all:

```
Play            V   S  Turn 2                 Turn 3      Turn 4          Turn 5      Turn 6
1C FLOR 19      -17 69 L8 SOSSES 26 (BNO +10)
9K ROLF 19      -20 66 O7 SOSSES 29 (BNO +10)
15A LO 33       -22 64 9I (Z)OS 25            7K (V)OR 7  5C (KITH)ES 17  F7 F(U)B 8  | L6 S(O)S 26 (N +2)
F7 B(U)FO 11    -23 63 L8 SOSSES 26 (LNOR +8)
N4 LOOF 11      -25 61 L8 SOSSES 26 (BNR +10)
11H (E)NROL 10  -26 60 M9 SOSSES 20 (BFO +16)
```

## Note on the dropped commit

This branch previously carried `3a19b462`, which widened the per-root-move aspiration fail-low toward `alpha_orig` instead of the multi-PV-raised `alpha`, aiming at the same goal (exact values for non-best root moves) from inside the main search. It has been dropped: it removes the widening loop's early exit, so once a good root move raises alpha, fail-lows that were previously accepted as bounds get re-searched with near-full windows. Measured four ways on the same harness (means of 2 reps, `-threads 4`):

| cell | base | alpha_orig | this PR | both | alpha_orig vs base |
| --- | --- | --- | --- | --- | --- |
| A p5 K1 | 12.34 | 13.27 | 12.39 | 13.29 | 1.1x |
| A p5 K8 | 12.28 | 13.28 | 12.82 | 13.56 | 1.1x |
| A p6 K1 | 25.40 | 27.79 | 25.38 | 27.86 | 1.1x |
| A p6 K8 | 25.88 | 27.56 | 25.12 | 27.60 | 1.1x |
| D p5 K1 | 0.12 | 8.24 | 0.12 | 8.16 | **68x** |
| D p5 K8 | 0.79 | 8.19 | 0.92 | 8.25 | **10x** |
| D p7 K1 | 0.15 | 50.67 | 0.15 | 50.97 | **344x** |
| D p7 K8 | 3.51 | 49.54 | 3.47 | 50.09 | **14x** |

The cost lands on `-etopk 1` too, where multi-PV is not active, and all four builds report identical values on A and D. Resolving the K displayed lines after the search is strictly cheaper and delivers the exact values that change was reaching for. The commit is preserved locally on `aspiration-alpha-orig-archive` if it is wanted separately.

## Checks

cppcheck, clang-tidy-18, format.py and find_circ_deps.py clean; `magpie_test endgame` and `magpie_test endgameinject` pass in release; a dev ASAN/UBSAN build solves `-eplies 4 -etopk 5` cleanly.

### DIFF
--- a/src/impl/endgame.c
+++ b/src/impl/endgame.c
@@ -2860,6 +2860,23 @@ static bool iterative_deepening_should_stop(EndgameCtx *solver) {
              THREAD_CONTROL_STATUS_USER_INTERRUPT;
 }
 
+// Zobrist hash of the root position as the search sees it: solving player to
+// move, no move played yet. Returns 0 when the TT is disabled.
+static uint64_t solver_root_hash_key(const EndgameCtx *solver,
+                                     const Game *game_copy) {
+  if (!solver->transposition_table_optim) {
+    return 0;
+  }
+  const Player *solving_player =
+      game_get_player(game_copy, solver->solving_player);
+  const Player *other_player =
+      game_get_player(game_copy, 1 - solver->solving_player);
+  return zobrist_calculate_hash(
+      solver->transposition_table->zobrist, game_get_board(game_copy),
+      player_get_rack(solving_player), player_get_rack(other_player), false,
+      game_get_consecutive_scoreless_turns(game_copy));
+}
+
 // Prepare solver->ext_game (lazy init or reset) then extend pv_line from TT.
 // Consolidates the repeated "prepare ext_game, then call pvline_extend_from_tt"
 // pattern used at all solver call sites.
@@ -2977,18 +2994,8 @@ void iterative_deepening(EndgameCtxWorker *worker, int plies) {
   }
   assert(worker->small_move_arena->size == 0); // make sure arena is empty.
 
-  uint64_t initial_hash_key = 0;
-  if (worker->solver->transposition_table_optim) {
-    const Player *solving_player =
-        game_get_player(worker->game_copy, worker->solver->solving_player);
-    const Player *other_player =
-        game_get_player(worker->game_copy, 1 - worker->solver->solving_player);
-    initial_hash_key = zobrist_calculate_hash(
-        worker->solver->transposition_table->zobrist,
-        game_get_board(worker->game_copy), player_get_rack(solving_player),
-        player_get_rack(other_player), false,
-        game_get_consecutive_scoreless_turns(worker->game_copy));
-  }
+  const uint64_t initial_hash_key =
+      solver_root_hash_key(worker->solver, worker->game_copy);
 
   // Half the threads use stuck-tile-aware root ordering (build chains +
   // conservation), the other half use score-based ordering for diversity.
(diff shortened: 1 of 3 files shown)


jvc56:
This change seem to make endgames way too slow, even for -etopk 1

src/impl/endgame.c
```diff
@@ -3387,22 +3585,38 @@ static int extract_multi_pvs(EndgameCtx *solver, EndgameCtxWorker *best_worker,
...
+  const bool resolve_exactly =
+      depth > 0 && !solver->first_win_optim && !solver->initial_window_optim;
+  if (resolve_exactly) {
+    filled = resolve_multi_pv_lines(solver, best_worker, depth, multi_pvs, k,
+                                    n_root, &next_candidate);
+  }
+
+  // Fallback for any slot the exact pass did not reach (interrupted, or
+  // disabled above): one root move plus a TT probe and greedy playout.
+  for (int r = filled; r < k && next_candidate < n_root;
```
olaugh: i know this is a rewrite of code that already had these r and k variable names, but let's give them more descriptive names now

src/impl/endgame.c
```diff
@@ -3357,13 +3364,202 @@ void *solver_worker_start(void *uncasted_solver_worker) {
...
+// other line would otherwise be the root move plus a transposition-table
+// probe and a greedy playout, which is not a solved variation. Because the
+// TT is warm from the main search, this re-search is cheap: only the PV
+// chain is re-expanded, every other node cuts off on a TT hit.
+//
+// Returns the root move's exact negamax value (spread-unadjusted, so callers
+// can compare with SmallMove estimated values), or ABDADA_INTERRUPTED if the
+// search was stopped. On success, out_pv holds the full line with
+// negamax_depth set to the number of exactly searched moves.
+static int32_t resolve_root_move_pv(const EndgameCtx *solver,
```
olaugh, line 3391: roughly reimplements lines 2576-2642, can we use a shared helper instead?

olaugh:
## Review: hands-on validation + code findings

### Part 1 — Hands-on validation (feature tried on real endgames)

Both the PR head (e944a7c0) and its base (0b596f78) were built as release binaries and run on positions taken from the existing endgame tests. Summary: **the feature works as advertised on every position tried, and it fixes a real duplicate-line bug that the base binary exhibits on both test positions.**

**1) Standard out-in-two endgame** (NWL20, BGIV vs DEHILOR, 384–389, `-eplies 4 -etopk 5 -threads 1`)

Base output had two defects the PR fixes:
- Lines 2–5 showed the `|` separator immediately after the first move, i.e. only the top line's continuation was actually solved.
- Line 3 was a **duplicate** of line 1 (`13K B(IT)` listed twice) — the stale `solver->principal_variation` de-dup bug described in the PR. The duplicate crowded out a distinct third-best move.

PR output: five distinct moves, every line solved to the end of the game:

```
13K B(IT) 9   V=11 S=6
1G VIG(A) 24  V=11 S=6
G7 B(EAU) 9   V=11 S=6   <- surfaced only after the duplicate was fixed
14N B(A) 8    V=10 S=5
7I (WE)B 8    V=10 S=5
```

Independent verification: for each of the five lines I constructed the child CGP by hand (first move placed, score updated, opponent to move) and solved each child position separately. Every child's final spread was exactly the negation of its parent line's S (−6, −6, −6, −5, −5), which is the invariant an exact solve must satisfy. Line arithmetic also verifies (e.g. line 1: 9 − 30 + 24 + 8 = 11 = V; spread −5 + 11 = 6 = S).

**2) Eldar v. Stick endgame** (CSW21, AEEIRUW/V, 410–409, `-eplies 5 -etopk 5 -threads 1`)

Base again produced a duplicate: `11B A(L)` appeared as both line 1 and line 2, and lines 2–5 carried the `|` right after move 1 (unproven). The PR produced five distinct moves, all V=72 S=73 (consistent with the 410–409 spread: 1 + 72 = 73), including `F3 R(ASH) 7`, a line the base binary's duplicate hid entirely. In the PR output the `|` appears only after ply 5 — exactly where the 5-ply proof depth ends, as designed.

**3) Thread consistency:** the same Eldar position at `-threads 4` returned the identical move set and identical values (ordering among ties differs, which the PR itself documents as expected).

**4) Time-limit behavior:** `-eplies 7 -etlim 5 -etopk 5` on the Eldar position returned in ~4s at depth 5 with all five lines present and values intact — clean degradation, no hang, no crash. (But see finding 1 below: the budget accounting for the resolve pass has a real bug in scripted contexts.)

**5) Test suite:** `./bin/magpie_test endgame` (release build, BOARD_DIM=15) passes on the PR head, exit 0.

### Part 2 — Code review findings (verified against PR-head source)

Correctness bugs, ranked by severity. Verdicts: CONFIRMED = reproducible from the code with a concrete failure path; PLAUSIBLE = credible but narrow/rare. Line numbers refer to the PR head.

1. **CONFIRMED — Resolve pass can overrun the time budget by up to the entire budget** (`resolve_multi_pv_lines`, ~line 3496). `remaining = hard_time_limit - endgame_results_get_seconds_elapsed(results)`, but `seconds_elapsed` is only written at reset (to 0) and inside `endgame_results_update_display_data` (endgame_results.c:234), whose only callers are the status renders in src/str/endgame_string.c:188/328. A scripted time-limited solve that never polls status computes `remaining = hard_time_limit - 0` and grants the resolve pass a fresh full budget. Non-PEG callers pass `external_deadline_ns = 0` (config.c:3200, play_chooser.c:352), so nothing else bounds it. This contradicts the PR's "gets only the unspent remainder" contract. Fix: read the live timer (`ctimer_elapsed_seconds(&results->timer)`) or capture elapsed at worker join.

2. **CONFIRMED — Promotion-scan early break assumes a sorted root array; the assumption is violated exactly when the final depth was aborted** (~3518). The abort path breaks out at 3233–3236 *before* the end-of-depth qsort (3242), and the root loop overwrites `estimated_value` in place per move (2773–2776), leaving a mixed-depth, unsorted arena. The break `estimate <= multi_pvs[k-1].score` then stops the scan on an out-of-order low value, and a later rightful top-K line is silently never resolved — the "displayed set is certain" guarantee fails precisely in the interrupted case. (Aborted-depth overwrites are also not valid upper bounds for the completed-depth comparison.) Fix: re-sort (or scan fully instead of breaking) before the promotion loop; this also closes finding 5.

3. **CONFIRMED — Display lock held across the entire resolve pass.** `endgame_solve` takes `endgame_results_lock(results, ENDGAME_RESULT_DISPLAY)` at 3804 before `extract_multi_pvs` at 3832. Pre-PR extraction was a handful of TT probes; now it is real negamax work — unbounded for untimed solves — and status is already FINISHED (3799), so a UI/API thread polling status blocks for the whole pass, indistinguishable from a hang.

4. **CONFIRMED — `-ttfraction 0` turns the pass into a cold full-window re-search per displayed line** (~3595). `resolve_exactly` checks only `depth > 0 && !first_win_optim && !initial_window_optim` — no TT gate — while every TT probe is behind `transposition_table_optim`. ttfraction 0 is accepted config and leaves the TT NULL, so the "warm TT makes this cheap" premise fails and cost multiplies by up to ~(1+K)×, unbounded when untimed. Fix: add `&& solver->transposition_table_optim` to the gate.

5. **CONFIRMED (narrow) — Actual-move analysis plus a top-value tie can drop a tied-best line.** `force_actual_move_to_front` (3244) and the de-dup swap compose cleanly only when the PV move lands in qsort slot 0; on an unstable-qsort tie they strand the low-valued actual move early and the tied-best move mid-array, so the promotion break fires on the actual move's exact low value and the tied-best line is never displayed.

6. **PLAUSIBLE — Slot-0 pinning is never re-checked** (`insert_resolved_pv`, 3457; `multi_pvs[0]` copied from results BEST at 3827). BEST is only overwritten on strictly-greater depth (endgame_results.c:238–245), and the resolve pass re-searches against the final warm TT (which can hold entries from a partially-searched deeper depth), so under multithreaded instability a resolved line's exact score can exceed pinned slot 0 — line 2 displayed as beating line 1, silently. A one-line guard (assert, or swap into slot 0) closes it.

7. **PLAUSIBLE — The set-certainty contract overstates what fail-soft estimates guarantee.** Non-top-K root estimates come from `alpha = MAX(alpha_orig, kth)` searches whose TT cutoffs (`ttentry_depth >= depth`; TT_UPPER entries from other threads' windows) make them non-rigorous upper bounds. Rare and silent; a comment softening the claim is probably the right fix, not code.

**Minor / latent:**
- Resolve-pass children satisfy `is_ply2` when `in_first_root_move` was left true, so `ply2_moves_total/completed` progress atomics keep churning after the search officially ended (3506). Cosmetic.
- The `remaining <= 0` early return (3499) skips zeroing `depth_deadline_ns`, unlike every other exit path. Latent only — `endgame_ctx_reset` re-zeroes it.

**Verified clean:** braces, include hygiene, const-correctness of the new functions, `solver_root_hash_key` de-duplication. `-etopk 1` is genuinely untouched (`extract_multi_pvs` is gated on `num_top_moves > 1`), and the `multi_pvs[0]` de-dup fix is correct — `solver->principal_variation` is only set by `endgame_solve_inline`, so the old threaded-path comparison was stale (and Part 1 shows the resulting duplicate on both test positions).

**Refuted during verification** (not findings): FINISHED-before-resolve status ordering (per-line fallback is visible via `|`, and pre-PR behavior was strictly worse); `search_complete` clear/re-set (workers are joined; safe).

**Bottom line:** the feature delivers what the PR promises on real endgames and fixes a user-visible duplicate-line bug, but findings 1–4 are worth fixing before merge — 1 and 2 break the PR's own contracts exactly in the time-limited/interrupted cases it calls out, and 3 can present as a UI hang.

—

*This review was written by Claude and posted from the account owner's login at their request.*

— Claude

jvc56:
Thanks for the thorough review — all nine comments are addressed in fab656c0.

## Inline comments

**`r` / `k` naming.** Renamed across the three multi-PV functions: `k` → `num_lines`, `n_root` → `num_root_moves`, the de-dup scan's `r` → `root_idx`, the fallback loop's `r` → `slot_idx`, `cand_idx` → `candidate_idx`. Also dropped the stale comment claiming the root moves are "already sorted … from the final qsort in `iterative_deepening`" — finding 2 shows that claim is exactly what was false.

**Shared helper for lines 2576-2642.** Added `play_small_move_and_hash(worker, small_move, node_key, undo)` above `abdada_negamax`: outplay detection, `small_move_to_move`, the empty leftover rack for outplays, the play, and `zobrist_add_move`, returning the child key. Both the negamax move loop and `resolve_root_move_pv` call it (~75 duplicated lines gone), and the explanation of *why* outplays need an empty leftover rack now lives in one place. The forced-pass fast path at ~2389 was deliberately left out: it uses `pass_undos[depth]` rather than `move_undos[undo_index]`, and a pass can never be an outplay.

## Findings

**1 — time budget overrun.** Fixed as suggested, by reading the live timer. Added `endgame_results_get_timer_elapsed_seconds()`, which returns `ctimer_elapsed_seconds(&results->timer)`; since `endgame_solve` stops the timer right after the join and before extraction, that *is* "elapsed at worker join", with no new state to keep in sync. `seconds_elapsed` and its getter are untouched for the status renderers.

Verified with `-eplies 8 -etlim 5 -etopk 8 -threads 1` on the Eldar position, driven over a pipe with no status polling: base computes `remaining = 5 - 0 = 5s` and resolves all eight lines on budget it does not have; the new binary computes `remaining = 5 - 9.4 < 0`, returns immediately, and the TT fallback fills lines 2–8.

**2 — promotion-scan break assumes a sorted array.** Fixed as suggested, by re-sorting: `extract_multi_pvs` now sorts `root_moves[1 .. n)` right after pinning the displayed best into slot 0, so the invariant the break relies on holds by construction. `resolve_multi_pv_lines`'s header states the requirement and names the function that establishes it. (Sorting the *tail* after the pin swap matters — sorting the whole array first and then swapping would move the highest element to wherever the pinned move came from, re-introducing the same out-of-order element.)

**3 — display lock held across the pass.** Fixed. `endgame_solve` picks `best_thread` unlocked (workers are joined), builds the lines into a locally allocated `PVLine` buffer with the lock *not* held — this is where the resolve pass runs — and then takes `ENDGAME_RESULT_DISPLAY` only to `memcpy` the finished lines in and set `num_pvs`. A poller now blocks for a `memcpy`, and never sees a half-built array. No new sharing: `solver_pvline_extend_from_tt` uses `solver->ext_game`, not the results' `ext_game`.

**4 — `-ttfraction 0`.** Fixed exactly as suggested: `&& solver->transposition_table_optim` added to the gate. Worth flagging the consequence — with the TT off, lines 2..K now show a bare root move with no continuation, because the fallback's greedy playout is itself TT-gated. That is the pre-PR `-ttfraction 0` display, and the alternative was up to (1+K)× the whole search, unbounded when untimed.

**5 — actual-move plus a top-value tie.** Closed by the fix for 2, as you anticipated. The re-sort runs after both `force_actual_move_to_front`'s effect and the de-dup swap, so a low-valued actual move can no longer sit ahead of a tied-best move. The re-sort's comment names both causes so the coupling is not re-broken later.

**6 — slot-0 pinning.** Fixed by swapping into slot 0 — one character: `insert_resolved_pv`'s sift-up bound `slot_idx > 1` → `slot_idx > 0`. The condition already required a *strictly* greater score, so a tie still cannot displace the search's own best line, while a line that genuinely beats it takes line 1. I preferred the swap to an assert since the condition is reachable in release builds.

**7 — set-certainty claim.** Comment only, as you suggested. The header no longer says the set "is certain"; it says the estimates are upper bounds in practice rather than rigorously, names the reason (a fail-soft estimate can be cut off by a TT entry stored from another thread's window), and adds the caveat this review surfaced — a candidate scanned after an aborted final depth carries a value from that partial depth. I'll soften the PR description the same way.

**Minor — ply-2 atomics.** `resolve_multi_pv_lines` sets `in_first_root_move = false` before the loop, so the pass's children no longer satisfy `is_ply2`.

**Minor — `remaining <= 0` early return.** Now zeroes `depth_deadline_ns` like every other exit path.

## Verification

Output is unchanged where it should be. Single-threaded (ties make `-threads 4` nondeterministic), `-ttfraction 0.02`, full table diffed with the timing line stripped: base vs new is **identical** on `BGIV/DEHILOR` NWL20 at 4 plies and `AEEIRUW/V` CSW21 at 5 plies, each at `-etopk 1`, `5` and `8`. The only behaviour changes are the two the findings asked for, and each appears only in its own configuration.

`magpie_test endgame`, `endgameinject`, `endgameoutplay`, `endgamefirstwin`, `eldar_v`, `sim`, `peg`, `pegtopkall`, `playchooser`, `cmdapi`, `autoplay`, `analyze` and the rest of the default table pass individually. (Running the entire default table in one invocation aborts after `Running test sim` on my machine — the unmodified base binary aborts at the identical point, so it is pre-existing and unrelated.)

Dev build (`-O0`, ASAN/UBSAN/LSAN, `-Werror`) compiles clean and solves `-eplies 3 -etopk 5 -threads 2` with no sanitizer output, which exercises the new malloc/memcpy/free path from finding 3. `format.py`, `cppcheck.sh` (2.17.1), `tidy.sh clang-tidy-18` and `find_circ_deps.py` are all clean.

olaugh:
Re-review of fab656c0. All nine findings from the previous review are verified fixed in the code, and the fix commit introduces no new issues I could find. Tested against real endgames on release builds (PR head vs. latest main, details below). **Ready to merge.**

## Fix verification

1. **Time budget** — `endgame_results_get_timer_elapsed_seconds` is a live `ctimer` read, and `endgame_results_stop_ctimer` runs before extraction, so the pass sees the true elapsed time (a stopped timer returns the frozen total). The `remaining <= 0` early return now zeroes `depth_deadline_ns`.
2. **Sort invariant** — the tail re-sort runs after both the de-dup swap and `force_actual_move_to_front`'s effect, uses the descending comparator, and correctly excludes pinned slot 0. The reply's note that sorting-then-swapping would re-break the invariant is right; the order implemented avoids it. Closes finding 5 as anticipated.
3. **Display lock** — lines are built into a local buffer unlocked and published as a `memcpy` under the lock, with no early return between the malloc and free. The now-unlocked read of the BEST pvline is safe: workers are joined, and the only concurrent accessor (`endgame_results_update_display_data`, called by the status renderers) only *reads* `best_pv_data`.
4. **TT gate** — `transposition_table_optim` added to `resolve_exactly`; verified behaviorally below.
5. Closed by 2.
6. **Slot-0 sift** — bound relaxed to `slot_idx > 0` with the strict inequality preserved, so a tie cannot displace the search's best line but a strictly better resolved line takes slot 0.
7. **Comment softened** as suggested.
- Both minors fixed (`in_first_root_move = false` before the pass; `depth_deadline_ns` zeroed on the early return).
- `play_small_move_and_hash` is a faithful extraction of both duplicated blocks. The loop's live-progress atomics read `small_move->tiny_move` directly (not `spare_move`), so the reordering the helper introduces is harmless.

## Endgame testing

Release builds, cold private TT per solve, data shared via symlink; base = latest main (0b596f78).

- **Test suites**: `endgame`, `endgameinject`, `endgameoutplay`, `endgamefirstwin`, `eldar_v`, `cmdapi`, `playchooser` all pass on the PR build.
- **Eldar `AEEIRUW/V` CSW21, 5 plies, `-etopk 8 -threads 4`**: every displayed line is a solved variation — the `|` separator sits at the depth-5 boundary on all 8 lines, values descend 72,72,72,72,72,71,71,71. The base binary on the same position shows the old one-exact-move lines **and a literal duplicate**: `15F (D)U(M)` appears as both line 1 and line 3, confirming the stale-PV de-dup bug on the threaded path is user-visible and that the `multi_pvs[0]` fix repairs it.
- **`-etopk 1` identity**: byte-identical to base at `-threads 1` on both the Eldar and SPORRAN positions (timing line stripped).
- **Time budget**: `-etlim 5` solves driven over a pipe with no status polling — exactly finding 1's failure scenario — finished in 3.84s (`-threads 4`, depth 5 completed, all 8 lines resolved from the leftover budget) and 1.44s (`-threads 1`, depth 4). No overrun in either case.
- **`-ttfraction 0`, `-etopk 5`**: byte-identical output and timing to base (0.13s vs 0.10s wall including startup) — the pass disables cleanly instead of becoming a cold re-search, and lines 2..K show the pre-PR bare-root-move display as documented.
- **Overhead**: 7-ply CSW24 `BFLNOOR/EOSSSS` at `-etopk 5 -threads 4` went from ~1.07s to ~1.21s mean-of-2 wall, while upgrading all 5 lines to fully solved variations (several played out to game end with rack-bonus annotations).

—

*This review was written by Claude and posted from the account owner's login at their request.*

— Claude

olaugh (approved):

PR #660, opened 2026-09-05
Reject racks larger than RACK_SIZE from CGP and GCG

### LINKED ISSUE #658: CGP accepts racks larger than RACK_SIZE, causing UB and a SEGV in move generation
## Summary

The CGP rack path validates that the requested tiles are *available in the bag*, but never checks the rack against `RACK_SIZE`. An over-full rack loads cleanly and then causes undefined behaviour and a SEGV in move generation.

Found while reviewing #615. It is **not** caused by that PR — reproduced on `main` at `2fb31c5c`.

## Reproduction

Over-full racks are accepted, while an unavailable-tile rack is correctly rejected:

```
$ printf 'set -lex CSW21 -wmp true\ncgp 15/15/15/15/15/15/15/15/15/15/15/15/15/15/15 <RACK>/ 0/0 0\n' | ./bin/magpie

rack ABCDEFG            ( 7 tiles) -> accepted
rack ABCDEFGHIJ         (10 tiles) -> accepted
rack ABCDEFGHIJKLMNOPQR (18 tiles) -> accepted
rack JJJ                ( 3 tiles) -> (error 51) rack not available in the bag for player 1: JJJ
```

Then generating moves on that position, dev build (ASan/UBSan):

```c
Config *config = config_create_or_die("set -lex CSW21 -wmp true -numplays 15");
load_and_exec_config_or_die(
    config, "cgp 15/15/15/15/15/15/15/15/15/15/15/15/15/15/15 "
            "ABCDEFGHIJKLMNOPQR/ 0/0 0");
Game *game = config_get_game(config);
MoveList *list = move_list_create(100);
const MoveGenArgs args = {.game = game, .move_list = list,
                          .eq_margin_movegen = 0,
                          .target_equity = EQUITY_MAX_VALUE,
                          .target_leave_size_for_exchange_cutoff = UNSET_LEAVE_SIZE};
generate_moves_for_game(&args);   // <-- crashes
```

Output:

```
OVERFULL rack tiles=18 (RACK_SIZE=7)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/impl/move_gen.c:2726:7
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/impl/../ent/leave_map.h:50:9
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/impl/../ent/leave_map.h:65:3
AddressSanitizer:DEADLYSIGNAL
==12782==ERROR: AddressSanitizer: SEGV on unknown address 0x000113478e38
SUMMARY: AddressSanitizer: SEGV leave_map.h:65 in leave_map_set_current_value
```

## Cause

`draw_rack_string_from_bag` (`src/impl/gameplay.c`) — the function `parse_cgp_racks_with_string_splitter` uses — calls `rack_set_to_string`, which is bounded by `MAX_RACK_SIZE = 100`, not `RACK_SIZE = 7`:

```c
int number_of_letters_set = rack_set_to_string(ld, &player_rack_copy, rack_string);
if (number_of_letters_set != -1) {
  if (!rack_is_drawable(game, player_index, &player_rack_copy)) {
    number_of_letters_set = -2;      // only "are the tiles in the bag?"
  } else {
    draw_rack_from_bag(game, player_index, &player_rack_copy);
  }
}
```

There is no size check, unlike `load_rack_or_push_to_error_stack` in `src/impl/config.c`, which does enforce it:

```c
} else if (num_mls > RACK_SIZE) {
  error_stack_push(error_stack, error_code,
      get_formatted_string("rack '%s' has %d tiles which exceeds the maximum rack size of %d",
                           rack_str, num_mls, RACK_SIZE));
}
```

Downstream, several structures are sized off `RACK_SIZE` and are indexed by the rack size, so an over-full rack runs past them — the leave map is simply the first to fault. In `src/impl/wmp_move_gen.h`: `count_by_size[RACK_SIZE + 1]`, `nonplaythrough_has_word_of_length[RACK_SIZE + 1]`, `nonplaythrough_best_leave_values[RACK_SIZE + 1]`, `nonplaythrough_infos[1 << RACK_SIZE]`, and the `RACK_SIZE + 1` entry `subracks_combination_offsets` table.

`gcg.c:399` looks like it may have the same gap — `rack_set_to_string(...) > 0` checks only the lower bound.

## Suggested fix

Reject `number_of_letters_set > RACK_SIZE` in `draw_rack_string_from_bag` (a third negative return, or routing the CGP path through the validated helper), so the error surfaces as a normal CGP parse error rather than as memory unsafety.

Found during an adversarial review of #615.

olaugh:
Fixes #658. Covers both entry points for an externally supplied rack: CGP and GCG.

## Problem

The CGP rack path checks that the requested tiles are *available in the bag*, but never checks the rack against `RACK_SIZE`. `draw_rack_string_from_bag` calls `rack_set_to_string`, which is bounded by `MAX_RACK_SIZE = 100`, so an 18-tile rack parses and gets drawn.

That is not merely wrong, it is memory-unsafe: move generation indexes several `RACK_SIZE`-sized arrays by the rack size (`count_by_size[RACK_SIZE + 1]`, `nonplaythrough_has_word_of_length[RACK_SIZE + 1]`, `nonplaythrough_best_leave_values[RACK_SIZE + 1]`, the `RACK_SIZE + 1` entry `subracks_combination_offsets`, and `nonplaythrough_infos[1 << RACK_SIZE]`). On `main`, generating moves from such a position gives UBSan errors and an ASan SEGV in `leave_map_set_current_value` — the leave map just happens to fault first.

## Fix

Reject the rack at parse time, before drawing from the bag.

- `draw_rack_string_from_bag` gains a `-3` return for "more than `RACK_SIZE` tiles", checked before `rack_is_drawable`.
- `parse_cgp_racks_with_string_splitter` surfaces it as a normal CGP parse error via a new `ERROR_STATUS_CGP_PARSE_RACK_TOO_MANY_LETTERS`.

The new status is appended to the end of the CGP group. Tests reference these by enum symbol and there is no parallel string table keyed on the numeric value, so appending renumbers later statuses harmlessly.

`draw_rack_string_from_bag` has exactly one production caller (`cgp.c`); the test callers ignore the return value and all pass racks within `RACK_SIZE`.

## Behaviour

```
rack ABCDEFG            ( 7 tiles) -> accepted
rack ABCDEFGH           ( 8 tiles) -> (error 52) rack for player 1 exceeds the maximum rack size of 7: ABCDEFGH
rack ABCDEFGHIJKLMNOPQR (18 tiles) -> (error 52) ...
rack JJJ                ( 3 tiles) -> (error 51) rack not available in the bag for player 1: JJJ   [unchanged]
rack <empty>            ( 0 tiles) -> accepted                                                     [unchanged]
```

## Tests

Added to the existing CGP failure cases in `game_test.c`: an 8-tile rack (one over the limit), an 18-tile rack, an over-full rack on player 2 rather than player 1, and a positive case asserting a full 7-tile rack still loads. The tiles in these racks are all available in the bag, so the size check is the only thing rejecting them.

## GCG (second commit)

An adversarial pass on the first commit found that GCG loading is a second, independent route to the same crash. Widening the first rack of a clean fixture to 12 tiles loaded silently, replay put all 12 tiles on the live player rack, and generating moves hit the same UBSan sites in `leave_map.h`. Replay applies racks through `set_rack_from_bag_or_push_to_error_stack`, not the CGP wrapper the first commit bounded.

The second commit refuses the rack in that helper **before any game state is touched**, with a new `ERROR_STATUS_GCG_PARSE_RACK_TOO_MANY_LETTERS` and the same shape of message as the CGP error:

```
$ load wide.gcg        # first rack widened to 12 tiles
(error 103) rack of ABCDEEEGORSY for player 1 has 12 tiles which exceeds the maximum rack size of 7
```

It also adds `assert(rack_get_total_letters(rack_to_draw) <= RACK_SIZE)` in `draw_rack_from_bag`, the one function every such path goes through (CGP, GCG replay, known racks) — the debug-only backstop for any caller that forgets. It is deliberately *not* the primary check: that helper maps every `false` return to a "not in bag" message, which would misdiagnose a size error.

Regression test: GCG fixtures live in the downloaded `testdata/`, not the repo, so the test carries a minimal valid GCG inline (pragma block plus two exchange events) and substitutes the first rack — 7 tiles as the control, then 8 and 10, which must be refused. Verified to bite: with both the helper bound and the backstop removed, the 8-tile case fails on `status == ERROR_STATUS_GCG_PARSE_RACK_TOO_MANY_LETTERS`; with only the helper bound removed, the backstop assert fires instead.

## Not included

The GCG *parser* (`gcg.c:399`) still accepts any rack length at parse time; the bound above is applied where the rack becomes live, which covers every source of a `GameHistory`, not only files. Rejecting earlier in the parser as well would give a line-numbered error; that is a nicety, not a safety gap.

## Checks

Full dev suite (ASan/UBSan) passed on the first commit; the second commit was verified with `gcg`, `game`, `cgp` and `gameplay` under ASan plus the end-to-end CLI load above, and cppcheck 2.17.1, the circular-dependency check and `format.py` are clean. Local clang-tidy is not usable on this machine (Homebrew LLVM 21.1.3 fails to analyze most files against the macOS headers); CI's clang-tidy-18 is the real gate.


src/impl/gameplay.c
```diff
@@ -430,6 +434,7 @@ void draw_leave_from_bag(Bag *bag, int player_draw_index, Rack *rack_to_update,
 // Returns number of letters drawn on success
 // Returns -1 if the string was malformed.
 // Returns -2 if the tiles were not in the bag.
+// Returns -3 if the string held more than RACK_SIZE tiles.
```
jvc56: At this point we should probably make these constants in gameplay.h to improve readability.

jvc56 (approved):
LGTM! Just had one comment. Feel free to merge when that's wrapped up.

src/impl/gameplay.c
olaugh: Done in c894f0b0: `DRAW_RACK_STRING_MALFORMED`, `DRAW_RACK_STRING_NOT_IN_BAG` and `DRAW_RACK_STRING_TOO_MANY_LETTERS`, used by the function and the CGP call site.

They live in a new `src/def/gameplay_defs.h` rather than `gameplay.h`: `gameplay.c` doesn't include its own header, and adding the self-include doesn't compile because the header's prototypes have drifted from the definitions (`Game *` vs `const Game *` for `draw_rack_from_bag`, `draw_starting_racks`, `set_random_rack`, possibly more). Filed as #668 rather than widening this PR. Happy to move them once that's fixed.

Issue #661, opened 2026-09-05
RACK_SIZE=8: SmallMove silently drops the 8th tile, so the endgame solver returns wrong results


olaugh:
## Summary

`SmallMove` packs at most 7 tiles into its 64-bit `tiny_move`, and nothing enforces that. Once `RACK_SIZE=8` builds (#659), an 8-tile move is **silently truncated to 7 tiles**: the metadata says 8, the payload holds 7, and the decoder hands back a 7-tile move. The endgame solver therefore returns wrong plays and scores at `RACK_SIZE=8` with no error, no assertion, and no sanitizer report.

This is the documented limitation in `src/ent/move.h:38-39` ("It encodes up to 7 tiles in 64 bits, so it only supports RACK_SIZE <= 7") — but it is enforced by comment only, and the failure mode is silent rather than loud.

## Evidence

Round-trip probe at `RACK_SIZE=8`, dev build (ASan/UBSan), encoding 7 tiles (control) and then 8 tiles with `small_move_set_all`, decoding each on an empty board with `small_move_to_move`:

```
SMALLMOVE n=7 meta_tiles_played=7 invalid_bit=0 tiny=0x07185103081001c0 decoded_tiles_played=7 decoded_len=7 decoded=1,2,3,4,5,6,7,
SMALLMOVE n=8 meta_tiles_played=8 invalid_bit=0 tiny=0x07185103081001c0 decoded_tiles_played=7 decoded_len=7 decoded=1,2,3,4,5,6,7,
```

The 8-tile `tiny_move` is byte-identical to the 7-tile one. The eighth tile (`H`, machine letter 8) is gone. No sanitizer output — unsigned overflow is well defined, so nothing trips.

## Mechanism

The writer places tile *i* at bit `20 + 6*i` with no bound on the tile index (`src/ent/move.h`, `small_move_set_all`):

```c
int bts = 20; // start at bitshift of 20 for the first tile
for (int i = 0; i < play_length; i++) {
  ...
  move_code |= (val << bts);
  tidx++;
  bts += 6;
}
```

[...]

PR #663, opened 2026-09-05
Fix bit_rack_defs.h for RACK_SIZE 4-6 and 8; add rack-size tests and a RACK_SIZE=8 CI job

### LINKED ISSUE #659: bit_rack_defs.h: swallowed #elif/#endif directives break the build for RACK_SIZE 4, 5, 6 and 8
## Summary

`main` only builds for `RACK_SIZE` 2, 3 and 7. Values 4, 5, 6 and 8 fail to compile, because three preprocessor directives in `src/def/bit_rack_defs.h` were absorbed into the body of a line-continued macro.

Reproduced on `main` at `2fb31c5c`, before any of #615.

| RACK_SIZE | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|
| `make magpie BUILD=release RACK_SIZE=n` | OK | OK | **FAIL** | **FAIL** | **FAIL** | OK | **FAIL** |

## Cause

`BIT_RACK_COMBINATION_OFFSETS` is defined with backslash continuations, and in three places the continuation runs into the next directive, so the `#elif` / `#endif` becomes ordinary macro-body text instead of a directive.

`src/def/bit_rack_defs.h:43-45`:

```c
#define BIT_RACK_COMBINATION_OFFSETS                                           \
  BIT_RACK_OFFSET_0, BIT_RACK_OFFSET_1, BIT_RACK_OFFSET_2, BIT_RACK_OFFSET_3,  \
      BIT_RACK_OFFSET_4 #elif RACK_SIZE == 5
```

`src/def/bit_rack_defs.h:58-60`:

```c
#define BIT_RACK_COMBINATION_OFFSETS                                           \
  BIT_RACK_OFFSET_0, BIT_RACK_OFFSET_1, BIT_RACK_OFFSET_2, BIT_RACK_OFFSET_3,  \
      BIT_RACK_OFFSET_4, BIT_RACK_OFFSET_5 #elif RACK_SIZE == 6
```

`src/def/bit_rack_defs.h:108-111`:

```c
#define BIT_RACK_COMBINATION_OFFSETS                                           \
  BIT_RACK_OFFSET_0, BIT_RACK_OFFSET_1, BIT_RACK_OFFSET_2, BIT_RACK_OFFSET_3,  \
      BIT_RACK_OFFSET_4, BIT_RACK_OFFSET_5, BIT_RACK_OFFSET_6,                 \
      BIT_RACK_OFFSET_7, BIT_RACK_OFFSET_8 #endif
```

The surviving `#if` chain is therefore `2, 3, 4, 7, 8` — the `== 5` and `== 6` arms no longer exist as directives:

```
12:#if RACK_SIZE == 2
21:#elif RACK_SIZE == 3
32:#elif RACK_SIZE == 4
68:#elif RACK_SIZE == 7
89:#elif RACK_SIZE == 8
112:#endif
```

That produces two distinct failures:

- **RACK_SIZE 5 and 6** match no arm, so `BIT_RACK_COMBINATION_OFFSETS` is never defined:
  `src/impl/wmp_move_gen.h:120:5: error: use of undeclared identifier 'BIT_RACK_COMBINATION_OFFSETS'`
- **RACK_SIZE 4 and 8** do take an arm, but the macro body now contains directive text, so it is a syntax error at the point of use:
  `src/impl/wmp_move_gen.h:120:5: error: expected '}'`

Both errors point at `wmp_move_gen.h`, which is only where the macro is *used* — the defect is in `bit_rack_defs.h`.

This looks like `clang-format` joining the lines (`format.py` runs clang-format-20 with no config file), so a fix should also be resistant to reformatting — e.g. keeping the trailing macro token and the following directive on clearly separate lines, or wrapping the definitions so no continuation ends adjacent to a directive.

## Why it went unnoticed

CI varies `BOARD_DIM` (15 and 21) but never `RACK_SIZE`, so nothing exercises these arms:

```
.github/workflows/run-tests.yml:170:  make magpie_test BUILD=release BOARD_DIM=21
.github/workflows/run-tests.yml:176:  make magpie_test BOARD_DIM=21
```

A cheap compile-only matrix over the supported `RACK_SIZE` values would catch this class of breakage.

Found during an adversarial review of #615, which adds a `RACK_SIZE`-sized array and so made this dimension worth exercising.

### LINKED ISSUE #662: rack_encode has no RACK_SIZE bound: over-full racks corrupt EncodedRack and rack_decode overflows
## Summary

`rack_encode` packs a rack into an `EncodedRack` whose capacity is derived from `RACK_SIZE` — `RACK_SIZE` (letter, count) slots with `BITS_TO_REPRESENT(RACK_SIZE)`-bit counts — but never checks that the rack actually fits. Encoding a rack with more tiles than `RACK_SIZE` corrupts the packed array, and `rack_decode` then reads back garbage machine letters and indexes `rack->array[]` out of bounds.

Reproduced on untouched `main` (`2fb31c5c`) by building at a small rack size, where the existing test feeds it 7-letter racks:

```
$ make magpie_test RACK_SIZE=2 && ./bin/magpie_test rack
==44099==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 2
    #0 rack_add_letters rack.h:163
    #1 rack_decode encoded_rack.h:128
    #2 assert_rack_encode_decode rack_test.c:238
    #3 test_encoded_rack rack_test.c:253
```

Same result at `RACK_SIZE=3`. At `RACK_SIZE=4`, `5`, `6` the corruption is milder and shows up as the round-trip assertion failing instead (`racks_are_equal(rack1, rack2)` at `rack_test.c:239`). `bitrack` passes at all of these, so this is specific to the encoder.

## Why it matters at the default RACK_SIZE too

At `RACK_SIZE=7` the test happens to pass because its longest racks are exactly 7. The encoder is still unbounded: any rack with more than 7 tiles corrupts on encode. Until #660, CGP accepted such racks (#658); the GCG rack path (`gcg.c:399`, `rack_set_to_string(...) > 0`) has not been checked. This is the same class of bug as #658 — a `RACK_SIZE` precondition that is assumed rather than enforced — one layer further in.

## Sizing (`src/ent/encoded_rack.h`)

```c
#define BITS_PER_COUNT (uint8_t)BITS_TO_REPRESENT(RACK_SIZE)
#define BITS_PER_RACK (uint8_t)(BITS_PER_ML_AND_COUNT * RACK_SIZE)
```

and `rack_encode` walks every letter with a non-zero count, emitting a slot per distinct letter and the raw count, with no check that distinct letters ≤ `RACK_SIZE` or that each count fits in `BITS_PER_COUNT`.

## Suggested fix

Two independent parts:

1. In `rack_encode`, assert (or return failure on) `rack_get_total_letters(rack) <= RACK_SIZE` before packing. That turns a silent corruption into a diagnosable failure in dev builds at zero release cost, matching how the other `RACK_SIZE`-sized structures are being handled.
2. Make `test_encoded_rack` rack-size-aware — either derive its racks from `RACK_SIZE` or skip the cases longer than `RACK_SIZE` — so the `rack` test can join the reduced suite that runs for non-default rack sizes. Right now it cannot, which is why the `RACK_SIZE=8` CI job proposed for #659 leaves it out.

Found while adding `RACK_SIZE` coverage for #659.

olaugh:
Fixes #659. Fixes #662.

## Problem

`main` only built for `RACK_SIZE` 2, 3 and 7. In `src/def/bit_rack_defs.h`, three line-continued `BIT_RACK_COMBINATION_OFFSETS` macros ran straight into the next preprocessor directive, so `#elif RACK_SIZE == 5`, `#elif RACK_SIZE == 6` and the closing `#endif` became macro-body text. The surviving chain was `2, 3, 4, 7, 8`: 5 and 6 matched nothing (`undeclared identifier`), 4 and 8 took an arm whose macro contained directive text (`expected '}'`). Both errors pointed at `wmp_move_gen.h`, which is only where the macro is used.

Restoring the directives was not enough: the `RACK_SIZE == 6` branch was also **incomplete** — it defined `BIT_RACK_C_6_*` but none of its `BIT_RACK_OFFSET_*` and no `BIT_RACK_COMBINATION_OFFSETS`.

CI never varies `RACK_SIZE`, so none of this was caught, and once other sizes did build, two more `RACK_SIZE`-sized structures turned out to have no bounds check either (#662).

## Fix

**`bit_rack_defs.h`** — all seven branches (2–8) rewritten uniformly in the style of the one branch that had survived formatting: parenthesized, one directive per line, and the final line of every `BIT_RACK_COMBINATION_OFFSETS` free of a trailing backslash. `format.py --write` leaves the result unchanged, which is the real test of whether it survives the formatter. The missing `== 6` offsets are added, and an `#error` guard after the chain makes a future missing branch fail here, with a clear message, instead of at a use site.

**`encoded_rack.h`, `leave_map.h`** — `rack_encode` and `leave_map_init` now assert `rack_get_total_letters(rack) <= RACK_SIZE`. Both structures are sized by `RACK_SIZE` and were silently corrupted (and `rack_decode` then read out of bounds) by a larger rack. Debug-only: `cflags.release` sets `-DNDEBUG`.

`DEFAULT_RACK_SIZE = 7` is added to `rack_defs.h`, next to the existing `DEFAULT_BOARD_DIM`.

## Tests: a separate file for non-default rack sizes

The standard `rack_test.c` and `leave_map_test.c` are full of 7-tile literals (`"ABCDEFG"`, `"DDDIIUU"`, `"???????"`) and assume `DEFAULT_RACK_SIZE`; they are left exactly as they are. New **`test/rack_size_test.c`** holds tests written so that every expectation is derived from `RACK_SIZE`:

- **combination offsets** — expands `BIT_RACK_COMBINATION_OFFSETS` and asserts it is the prefix sums of C(RACK_SIZE, k) for k in [0, RACK_SIZE], with `RACK_SIZE + 1` entries, ending at `1 << RACK_SIZE`. Data-free.
- **BitRack counts** — a full rack with repeats and a blank round-trips through `bit_rack_create_from_rack`.
- **EncodedRack round trips** — every size 0..RACK_SIZE, for distinct letters, one repeated letter, `Z` repeated, blanks, and a mixed full rack.
- **LeaveMap indexing** — a full distinct rack starts at `(1 << RACK_SIZE) - 1` and letter *p* owns bit *p*; a full rack of one letter walks `(1 << (RACK_SIZE - k)) - 1` down to 0 as copies are taken and back up as they are added, with stored values reading back at the same indices.
- **CGP loading** — the integration check, since a CGP is where a `RACK_SIZE`-tile rack first enters the engine from outside. Builds an empty `BOARD_DIM` board with racks of exactly `RACK_SIZE` tiles (distinct letters, one repeated letter, both blanks plus letters), one tile short, and empty; asserts the loaded rack sizes, that the racks field comes back verbatim from `game_get_cgp`, and bag accounting relative to an empty-racks load. Then generates moves from the loaded position and asserts every move plays at most `RACK_SIZE` tiles and at least one tile placement exists. Rejection of a `RACK_SIZE + 1` rack is not asserted yet — that check is #660's and is not on this branch; it is a one-line addition once #660 lands.

Registered as `racksize` in the standard table, so it also runs in the default suite at `RACK_SIZE=7`. Before trusting the offsets check I verified it fails: shifting C(8,2) by one aborts on `offsets[size] == expected_offset`; dropping the last list entry aborts on the count; an offset pushed to 256 is rejected by the compiler (`-Werror` int→`uint8_t` narrowing).

## CI: `RACK_SIZE=8`, and only 8

New `unit-tests-rack8` job, cloned from `unit-tests-super`: release build + run, `make clean`, dev build + run, both at `RACK_SIZE=8`; added to `ci-summary`. 8 is the largest supported value and where a fixed-size array is tightest. 2–6 build and pass the same suite locally (table below) but are deliberately **not** in CI — not worth the extra compiles.

`test.c` gains a `RACK_SIZE != DEFAULT_RACK_SIZE` branch. With no arguments it runs the reduced suite — `racksize`, `bitrack`, `wmp`, `rit` (the last three are standard tests that happen to be rack-size-generic). Named tests still run at any rack size, so a single test can be debugged there. The default and super paths are unchanged. The endgame tests are left out because `SmallMove` encodes at most 7 tiles and silently truncates an 8th — #661, found while doing this.

## Verification

| RACK_SIZE | build | reduced suite (dev, ASan/UBSan) |
|---|---|---|
| 2 | ok | pass |
| 3 | ok | pass |
| 4 | ok | pass |
| 5 | ok | pass |
| 6 | ok | pass |
| 7 | ok | default path unchanged; `racksize`, `bitrack`, `rack`, `leavemap`, `game`, `gameplay`, `cgp`, `wmg` pass — `rack`/`leavemap` exercise the new asserts on legitimate 7-tile racks |
| 8 | ok | pass — dev **and** release, mirroring the CI job |

cppcheck 2.17.1 clean, circular-dependency check clean, `format.py` a no-op on the result. Local clang-tidy is not usable on this machine (Homebrew LLVM 21.1.3 fails to analyze most files against the macOS headers); CI's clang-tidy-18 is the gate.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -176,6 +176,45 @@ jobs:
           make magpie_test BOARD_DIM=21
           ./bin/magpie_test
 
+  unit-tests-rack8:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up GCC
+        uses: egor-tensin/setup-gcc@v1
+        with:
+          version: latest
+          platform: x64
+
+      - name: Get datafile cache
+        id: cache-datafiles
+        uses: actions/cache@v3
+        with:
+          path: |
+            data/*
+            testdata/*
+          key: downloaded-data-files-30
+
+      - name: Download data files and convert lexica
+        if: steps.cache-datafiles.outputs.cache-hit != 'true'
+        run: ./download_data.sh && ./convert_lexica.sh
+
+      # RACK_SIZE is a compile-time parameter that CI otherwise never varies.
+      # 8 is the largest supported value; 2-6 build too but are left to local
+      # runs (make magpie_test RACK_SIZE=n && ./bin/magpie_test) to keep CI
+      # cheap. The no-argument test run is the reduced rack-size-generic set.
+      - name: Build and run RACK_SIZE=8 release tests
+        run: |
+          make magpie_test BUILD=release RACK_SIZE=8
+          ./bin/magpie_test
+
+      - name: Build and run RACK_SIZE=8 dev tests
+        run: |
+          make clean
+          make magpie_test RACK_SIZE=8
+          ./bin/magpie_test
+
   wasm-tests:
     runs-on: ubuntu-latest
     timeout-minutes: 15
@@ -213,6 +252,7 @@ jobs:
       - unit-tests-no-pgo-release
       - unit-tests-dev
       - unit-tests-super
+      - unit-tests-rack8
       - wasm-tests
     runs-on: ubuntu-latest
     steps:
@@ -228,6 +268,7 @@ jobs:
           echo "| unit-tests-no-pgo-release | \`${{ needs.unit-tests-no-pgo-release.result }}\` |" >> $GITHUB_STEP_SUMMARY
           echo "| unit-tests-dev | \`${{ needs.unit-tests-dev.result }}\` |" >> $GITHUB_STEP_SUMMARY
           echo "| unit-tests-super | \`${{ needs.unit-tests-super.result }}\` |" >> $GITHUB_STEP_SUMMARY
+          echo "| unit-tests-rack8 | \`${{ needs.unit-tests-rack8.result }}\` |" >> $GITHUB_STEP_SUMMARY
           echo "| wasm-tests | \`${{ needs.wasm-tests.result }}\` |" >> $GITHUB_STEP_SUMMARY
 
       - name: Check for failures
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -206,7 +206,7 @@ jobs:
       # cheap. The no-argument test run is the reduced rack-size-generic set.
       - name: Build and run RACK_SIZE=8 release tests
         run: |
-          make magpie_test BUILD=release RACK_SIZE=8
+          make magpie_test BUILD=no_pgo_release RACK_SIZE=8
           ./bin/magpie_test
 
       - name: Build and run RACK_SIZE=8 dev tests
--- a/src/def/bit_rack_defs.h
+++ b/src/def/bit_rack_defs.h
@@ -9,62 +9,87 @@ enum {
   BIT_RACK_EXPECTED_SIZE_BYTES = 128 / 8
 };
 
+// BIT_RACK_COMBINATION_OFFSETS lists, for each subrack size k in
+// [0, RACK_SIZE], the index at which size-k subracks begin in a flat array of
+// all canonical subracks: the prefix sums of C(RACK_SIZE, k). The last offset
+// plus C(RACK_SIZE, RACK_SIZE) is therefore 1 << RACK_SIZE.
+//
+// Formatting hazard: these are multi-line macros directly followed by
+// preprocessor directives. A backslash on the LAST line of a macro body
+// swallows the next #elif/#endif into the macro, which silently deletes that
+// branch and breaks the build for other RACK_SIZE values (jvc56/MAGPIE#659).
+// Keep the final line of each BIT_RACK_COMBINATION_OFFSETS free of a trailing
+// backslash. The #error guard after the chain turns a missing branch into a
+// clear compile error, and test_combination_offsets in test/bit_rack_test.c
+// checks the values and the entry count at every RACK_SIZE.
 #if RACK_SIZE == 2
-#define BIT_RACK_C_2_0 1
-#define BIT_RACK_C_2_1 BIT_RACK_C_2_0 * 2 / 1
-#define BIT_RACK_C_2_2 BIT_RACK_C_2_0
-#define BIT_RACK_OFFSET_0 0
-#define BIT_RACK_OFFSET_1 BIT_RACK_OFFSET_0 + BIT_RACK_C_2_0
-#define BIT_RACK_OFFSET_2 BIT_RACK_OFFSET_1 + BIT_RACK_C_2_1
+#define BIT_RACK_C_2_0 (1)
+#define BIT_RACK_C_2_1 (BIT_RACK_C_2_0 * 2 / 1)
+#define BIT_RACK_C_2_2 (BIT_RACK_C_2_0)
+#define BIT_RACK_OFFSET_0 (0)
+#define BIT_RACK_OFFSET_1 (BIT_RACK_OFFSET_0 + BIT_RACK_C_2_0)
+#define BIT_RACK_OFFSET_2 (BIT_RACK_OFFSET_1 + BIT_RACK_C_2_1)
 #define BIT_RACK_COMBINATION_OFFSETS                                           \
   BIT_RACK_OFFSET_0, BIT_RACK_OFFSET_1, BIT_RACK_OFFSET_2
 #elif RACK_SIZE == 3
-#define BIT_RACK_C_3_0 1
-#define BIT_RACK_C_3_1 BIT_RACK_C_3_0 * 3 / 1
-#define BIT_RACK_C_3_2 BIT_RACK_C_3_1
-#define BIT_RACK_C_3_3 BIT_RACK_C_3_0
-#define BIT_RACK_OFFSET_0 0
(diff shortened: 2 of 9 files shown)


olaugh:
Rebased onto main past #616 (clean, 8 files) with one addition: #615's `test_nonplaythrough_subrack_enumeration` landed after this PR and spells its racks with seven tiles and its expected counts with `RACK_SIZE`-7 slots, so `RACK_SIZE` 4–6 had stopped building. The seven-tile cases are now behind `#if RACK_SIZE >= 7`; the empty and single-tile cases hold at every size and stay unguarded.

Verified on the rebase: dev build plus the reduced suite at `RACK_SIZE` 4, 5, 6 and 8, release build plus the reduced suite at 8 (mirroring the new CI job), and `racksize bitrack rack leavemap cgp wmg movegen` at 7; cppcheck, circular-deps and `format.py` clean.

For the record, not a regression: `wmg` at `RACK_SIZE=8` fails in `test_nonplaythrough_existence` (`best_leaves[leave_size] == leave_size` assumes a seven-tile rack) — one more of the 7-tile-literal tests this PR intentionally leaves out of the non-default suite.

PR #670, opened 2026-09-06
make release: find versioned clang, fail fast without it, else build plain

olaugh:
## Problem

`PGO_CC ?= clang` was handed to three sub-makes unchecked. On a machine without that exact name — gcc-only, or a distro that installs only `clang-18` — `make release` compiled half the tree and then died on the first object:

```
make[2]: clang: Command not found
make[2]: *** [Makefile:182: obj/no_pgo_release-b15-r7/src/str/move_string.o] Error 127
./run: line 69: ./bin/magpie: No such file or directory
```

## What this does

- **Finds the toolchain.** `PGO_CC` tries `clang`, then `clang-20` … `clang-15`; `LLVM_PROFDATA` tries the one matching a versioned clang first (`clang-18` → `llvm-profdata-18`), then the plain and versioned names, then `xcrun` on macOS. An explicit `PGO_CC=` / `LLVM_PROFDATA=` still wins.
- **Fails before compiling anything.** A `pgo_toolchain_check` prerequisite of `pgo_workload` refuses to start with no compiler, with a `PGO_CC` that is not clang (the instrumentation flags are clang's), or with no `llvm-profdata`, and names the variable to set.
- **`make release` without clang builds the plain optimized release** and says so, so the documented entry point — and `./run r` — still produce a fast binary on a gcc-only box.
- `run r` no longer launches `bin/magpie` after a failed build (`&&`).
- README / AGENTS.md describe the detection and the fallback.

## Validation (M4, macOS)

| path | result |
|---|---|
| `make release` (clang + `xcrun llvm-profdata` detected) | PGO build end to end, exit 0; `pgo_generate` / `pgo_use` / `pgo-data` present; binary loads CSW24 |
| `make release PGO_CC= LLVM_PROFDATA=` (no toolchain) | notice, then `make magpie BUILD=no_pgo_release`; `bin/magpie` built |
| `make release PGO_CC=clang-99` | `*** PGO_CC=clang-99 was not found on PATH.` — nothing compiled |
| `make release PGO_CC=<fake gcc>` | `*** PGO_CC=… is not clang; the PGO instrumentation flags are clang-specific.` |
| `make release LLVM_PROFDATA=` | `*** PGO needs llvm-profdata (…); pass LLVM_PROFDATA=<path>.` |

Not testable here: an Ubuntu box with only `clang-18` on PATH — that is the detection path this is for, and worth one run there.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,9 +26,11 @@ make libmagpie                 # shared library; API is src/impl/cmd_api.h
 make clean                     # remove build artifacts
 ```
 
-The local PGO build uses Clang and `llvm-profdata`, reuses an existing production
-CSW24 RIT (or creates it when missing), discards old profile data, and trains
-the current source on static autoplay. Experimental `pgo`, `pgo_sim`,
+The local PGO build uses Clang and `llvm-profdata`, found under their plain or
+distro-versioned names (`clang-18`, `llvm-profdata-18`); with no clang on PATH,
+`make release` builds the plain optimized release and says so. It reuses an
+existing production CSW24 RIT (or creates it when missing), discards old
+profile data, and trains the current source on static autoplay. Experimental `pgo`, `pgo_sim`,
 `pgo_peg`, and `pgo_eg` targets are also available. Each produces a
 `-march=native` binary for the build machine. Rerun the selected target after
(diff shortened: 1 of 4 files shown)


jvc56 (approved):
LGTM! Confirmed the fix on my machine.

PR #674, opened 2026-09-07
Do not add leaves to the WMP subrack bound under score sort

### LINKED ISSUE #673: WMP movegen: score-sorted within-x lists drop vertical plays, including ties for the best
With a WMP loaded, `-s score`, and the within-x record mode (`MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST`, `-r equity`), move generation drops some plays that score above the cutoff, including ties for the best score. The generator without a WMP returns the complete list for the same positions, and equity sort is complete on the same positions up to a 40-point margin, so this is specific to the WMP path under score sort.

## Reproduce

- `set -lex CSW21 -wmp true`
- position `DOUG_V_EMELY_CGP` from `test/test_constants.h`: `15/15/15/15/15/15/15/3WINDY7/15/15/15/15/15/15/15 ADEEGIL/AEILOUY 0/32 0`
- `MoveGenArgs`: `move_record_type = MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST`, `move_sort_type = MOVE_SORT_SCORE`, `eq_margin_movegen = int_to_equity(20)`, a move list of capacity 100000, `generate_moves()` directly (`generate_moves_for_game()` would replace the record type with the player's).

The best score is 22, so the cutoff is 2. `MOVE_RECORD_ALL` returns 692 plays scoring above 2; the within-x list has 691. The missing play is **E7 LID** (2 tiles through the I of WINDY, 4 points). With `-wmp false` the within-x list has all 692.

Other positions in the same set lose more: on one, a 40-point margin drops 11 of 106 above-cutoff plays; on another, a 5-point margin loses one of the plays tied for the top score (24 points), so the returned "best" set is wrong, not only the tail. Every play I saw dropped was vertical, with or without playthrough tiles.

## Where to look

`test_wmp_bounded_record_modes` in `test/wmp_move_gen_test.c` (from #612) compares the bounded record modes against the full list on 11 positions in both sort types; its within-x check is limited to equity sort because of this. Removing the `sort_type != MOVE_SORT_EQUITY` early return in `assert_bounded_record_modes_match_full_list` reproduces the failures under ASan (`./bin/magpie_test wmg`), and printing the plays the full list has that the within-x list lacks gives the list above.

Reproduces identically on the #614 base (`6847ade5`) and is unrelated to #612's shortcut (the differential in #612 shows the shortcut changes nothing); the within-x handling in the WMP path predates both.

olaugh:
Fixes #673.

## What this does

In the WMP path, `wordmap_gen_record_subrack` skips a subrack when the anchor's highest possible score plus the subrack's leave value cannot beat the record cutoff. That sum bounds **equity**, which is what the cutoff measures under equity sort. Under score sort the cutoff is a **score**, and a negative leave pulls the bound below plays the subrack can still make, so the subrack is skipped and its plays are never generated. It bites in whichever direction is generated second, once the cutoff has risen.

The leave now takes part in the bound only when the bag is not empty and the sort is by equity. The decision is made once per generation in `gen_load_position` (`wmp_prune_subracks_by_leave`), so the per-subrack test stays a single flag check, as it was. Under score sort nothing replaces it: the anchor-level check in `gen_record_scoring_plays` already bounded the score.

```c
  // move_gen.c, wordmap_gen_record_subrack
  if (gen->wmp_prune_subracks_by_leave) {
    const Equity leave_value = wmp_move_gen_get_leave_value(wgen, subrack_idx);
    if (better_play_has_been_found(gen, leave_value +
                                            anchor->highest_possible_score)) {
      return false;
    }
  }

  // gen_load_position
  gen->wmp_prune_subracks_by_leave =
      (gen->number_of_tiles_in_bag > 0) &&
      (gen->move_sort_type == MOVE_SORT_EQUITY);
```

## Where it came from

The original WMP move generator. The subrack prune was written in `b6a84ef6` (2025-09-11, "functions for wmp movegen in move_gen.c (not called)") with only the `bag > 0` guard, and went live the next day in `d8064509`. The non-WMP path had guarded its own leave-based prune (`record_exchange`) with `move_sort_type == MOVE_SORT_EQUITY` since early 2024; the WMP port omitted that guard. Later refactors of this loop (the masked-scan split in September 2026) moved the lines and kept the logic. So it is not a regression from an equity-sort optimization; it has been there for every score-sorted WMP generation since the WMP path shipped.

## What it affected

- `MOVE_RECORD_WITHIN_X_EQUITY_OF_BEST` under score sort: lists missing plays above the cutoff, including ties for the best (E7 LID for 4 on the WINDY position; a 24-point tie on another).
- `MOVE_RECORD_BEST` under score sort is exposed to the same skip whenever a better play scores less than |leave| above the running best and its anchor's bound is not slack enough to absorb it. And it did: over 400 score-sorted autoplay games `main`'s WMP generator diverges from the non-WMP generator's play, and the fixed one matches it move for move (differential below).
- Equity sort is unaffected: the bound is correct there, and the test below holds on `main` for it.

## Correctness

- **New test** `test_wmp_cutoff_modes_are_complete` (`wmg` suite): drives `MOVE_RECORD_BEST` and within-x lists (margins 0/5/20/40) through eleven CSW21 positions in both sort types and checks each against the full list — the best move carries the head's equity, and a within-x list holds every play strictly inside the margin, with equal-equity ties compared by membership. It fails on `main` under score sort (`cutoff_modes_count_above(within, cutoff) == cutoff_modes_count_above(full, cutoff)` on the first position) and passes with the fix.
- **Autoplay differential** against `main` (200 game pairs, CSW24 with WMP, WIT, RIT, seed 4242): with the WMP, WIT and RIT on, seed 4242. Score sort, `best`: the fixed generator's games are **identical to the generator without a WMP** (`-wmp false`, the reference path), while `main`'s WMP games diverge from it (9,230 vs 9,498 turns over the 400 games). Equity sort, `best`: WMP and non-WMP games identical on both, so equity sort is untouched. Score sort with `all -numplays 5` and with within-x: fixed vs `main` differ, as expected.
- ASan `wmg movegen shadow`; release `wmg movegen shadow sim witsweep witdiff`; cppcheck, circular-dependency check and `format.py` clean.

## Performance

No difference. In the benchmark workload (equity sort, bag not empty) the flag is true exactly when the old `bag > 0` test was true, so the pruning decisions are the same as on `main`; the only question was codegen, and it comes out even. M4 Mac mini, 8 workers, `no_pgo_release`, `simbench` 0.5 s/turn, WIT on, seeds 42/24301/1552 × 6 interleaved rounds, paired within-round geometric means, stratified-bootstrap 95% CIs, 18 pairs per row:

| configuration | fix vs `main` | 95% CI |
|---|---:|---:|
| RIT on | +0.03% | [-0.22%, +0.35%] |
| RIT off | +0.01% | [-0.09%, +0.11%] |

The flag exists because the alternative costs: writing the fix as `bag > 0 && sort == equity` per subrack, with no new field, measured **-1.44%** [-1.79%, -1.05%] against `main` in the same campaign. Where the flag sits in `MoveGen` makes no difference (a variant with it at the end of the struct is within 0.2% either way). The full codegen, layout and profile comparison is in a comment below.

## Related

- #612's `test_wmp_bounded_record_modes` limits its within-x check to equity sort because of this bug; once this merges that guard can go.

### DIFF
--- a/src/impl/move_gen.c
+++ b/src/impl/move_gen.c
@@ -1002,7 +1002,12 @@ static inline __attribute__((always_inline)) bool
 wordmap_gen_record_subrack(MoveGen *gen, const Anchor *anchor, int subrack_idx,
                            bool lazy) {
   WMPMoveGen *wgen = &gen->wmp_move_gen;
-  if (gen->number_of_tiles_in_bag > 0) {
+  // The anchor's score bound plus this subrack's leave bounds the equity of
+  // every play the subrack can make, so under equity sort a subrack whose
+  // bound cannot reach the cutoff is skipped. Under score sort the cutoff is
+  // a score, which the anchor-level check already bounded; adding a leave
+  // there would skip subracks that still hold plays above the cutoff.
+  if (gen->wmp_prune_subracks_by_leave) {
     const Equity leave_value = wmp_move_gen_get_leave_value(wgen, subrack_idx);
     if (better_play_has_been_found(gen, leave_value +
                                             anchor->highest_possible_score)) {
@@ -3148,6 +3153,8 @@ void gen_load_position(MoveGen *gen, const MoveGenArgs *args) {
 
   gen->bingo_bonus = game_get_bingo_bonus(game);
   gen->number_of_tiles_in_bag = bag_get_letters(game_get_bag(game));
+  gen->wmp_prune_subracks_by_leave = (gen->number_of_tiles_in_bag > 0) &&
+                                     (gen->move_sort_type == MOVE_SORT_EQUITY);
   gen->kwgs_are_shared = game_get_data_is_shared(game, PLAYERS_DATA_TYPE_KWG);
   gen->move_list = move_list;
   gen->cross_index =
--- a/src/impl/move_gen.h
+++ b/src/impl/move_gen.h
@@ -90,6 +90,10 @@ typedef struct MoveGen {
   int move_sort_type;
   move_record_t move_record_type;
   int number_of_tiles_in_bag;
+  // Whether a subrack's leave may take part in the bound that decides if the
+  // subrack can still beat the cutoff: only when the bag is not empty and
+  // the sort is by equity.
+  bool wmp_prune_subracks_by_leave;
   int player_index;
   Equity bingo_bonus;
   bool kwgs_are_shared;
--- a/test/wmp_move_gen_test.c
+++ b/test/wmp_move_gen_test.c
@@ -735,6 +735,161 @@ static void test_playthrough_moves_against_recursive(void) {
   config_destroy(reference);
 }
 
+// Under score sort the cutoff is a score, so a subrack's leave value must take
+// no part in the bound that decides whether the subrack can still beat the
+// cutoff (#673). This drives the WMP generator through the cutoff-based record
+// modes on a set of positions in both sort types and checks each result
+// against the full list: the best move carries the head's equity, and a
+// within-x list holds every play strictly inside the margin. Equal-equity
+// ties are compared by membership rather than position, since shadow lets the
+// bounded modes skip anchors that cannot beat the running cutoff and that
+// decides ties by anchor order.
+enum { CUTOFF_MODES_FULL_CAPACITY = 100000 };
+
+// A SortedMoveList borrows its Move objects from the MoveList it was built
+// from, so the two live and die together.
+typedef struct CutoffModesGeneration {
+  MoveList *move_list;
+  SortedMoveList *sorted;
+} CutoffModesGeneration;
+
+static CutoffModesGeneration
+cutoff_modes_generate(Game *game, move_record_t record_type,
+                      move_sort_t sort_type, Equity eq_margin, int capacity) {
+  CutoffModesGeneration generation;
+  generation.move_list = move_list_create(capacity);
+  const MoveGenArgs args = {
+      .game = game,
+      .move_list = generation.move_list,
+      .move_record_type = record_type,
+      .move_sort_type = sort_type,
+      .eq_margin_movegen = eq_margin,
+      .target_equity = EQUITY_MAX_VALUE,
+      .target_leave_size_for_exchange_cutoff = UNSET_LEAVE_SIZE,
+  };
+  // Not generate_moves_for_game(): that replaces the record type with the
+  // on-turn player's configured one.
+  generate_moves(&args);
+  generation.sorted = sorted_move_list_create(generation.move_list);
+  return generation;
+}
+
+static void cutoff_modes_generation_destroy(CutoffModesGeneration *generation) {
+  sorted_move_list_destroy(generation->sorted);
+  move_list_destroy(generation->move_list);
+}
+
+static bool cutoff_modes_moves_match(const Move *move_1, const Move *move_2) {
+  if (move_get_type(move_1) != move_get_type(move_2) ||
+      move_get_row_start(move_1) != move_get_row_start(move_2) ||
+      move_get_col_start(move_1) != move_get_col_start(move_2) ||
+      move_get_dir(move_1) != move_get_dir(move_2) ||
+      move_get_tiles_played(move_1) != move_get_tiles_played(move_2) ||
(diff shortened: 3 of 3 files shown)


olaugh:
For the record, the performance investigation that the description used to carry. The short version is in the description: no difference on this base; the per-subrack alternative costs 1.4%.

Neutral on the base this PR sits on (`main` with #614). In the benchmark workload (equity sort, bag not empty) the new flag is true exactly when the old `bag > 0` test was true, so every pruning decision is the same as on `main`; the numbers only measure codegen. M4 Mac mini, 8 workers, `no_pgo_release`, `simbench` 0.5 s/turn, WIT on, seeds 42/24301/1552 × 6 interleaved rounds, paired within-round geometric means, stratified-bootstrap 95% CIs, 18 pairs per row:

| configuration | fix vs `main` | 95% CI |
|---|---:|---:|
| RIT on | +0.03% | [-0.22%, +0.35%] |
| RIT off | +0.01% | [-0.09%, +0.11%] |

### The +0.3% that was there before, and where it went

Measured on the pre-#614 `main` (`94961e9a`, before the rebase), the same change read +0.32% [+0.24%, +0.42%] with the RIT on and +0.24% [+0.16%, +0.32%] with it off — small, but with intervals excluding zero and seeds agreeing within a tenth of a percent, so worth explaining rather than waving at. Three things were checked.

**Generated code.** On `main` the per-subrack test is `ldr w, [gen, #0x38]` (the `int` bag count), a compare and a branch, and the compiler keeps the bag count live across the loop — in the masked scan it spills it to the stack. With the flag it is `ldrb w, [gen, #0x3c]`, `cmp`, `b.ne`, and the bag count is loaded less often. The three functions that inline the subrack loop all shrink:

| function (instructions, release build) | main | fix | flag at end of struct | inline two-field test |
|---|---:|---:|---:|---:|
| `generate_moves` | 2403 | 2399 | 2400 | 2403 |
| `gen_record_scoring_plays_eager` | 614 | 609 | 610 | 621 |
| `wordmap_gen_forbidden_subracks` | 269 | 259 | 262 | 272 |
| loads of `number_of_tiles_in_bag` in those three | 12 | 8 | 8 | 12 |

The last two columns are the two alternatives built to separate causes: the same fix with the flag moved to the *end* of `MoveGen` (fits in tail padding, so no other field moves), and the fix written without a flag as `bag > 0 && sort == equity` per subrack (one more load and compare each time).

**Layout.** The flag sits at offset 60, after `number_of_tiles_in_bag`, which shifts 75 of the struct's 88 fields by 4–8 bytes and moves several hot ones across 64-byte-line boundaries (`wmp_move_gen`, `wit_len_lane`, `strip`, `shadow_perpendicular_additional_score` among them). The flag-at-end variant shifts nothing (same `sizeof`), so comparing it with the fix isolates the layout effect from the cheaper test.

**Profile** (`BUILD=profile`, inlining suppressed, one thread, RIT on): leaf shares move by fractions of a point, within sampling noise; `generate_moves`, which holds the inlined subrack loop, is −0.29 pp. Nothing outside movegen moves.

**The four variants in one paired campaign** on the current base (RIT on, same design, 18 pairs per row; a first campaign was discarded because I ran the disassembly and profile analysis on the same machine during it):

| variant vs `main` | delta | 95% CI |
|---|---:|---:|
| fix (flag after the bag count) | +0.03% | [-0.22%, +0.35%] |
| fix, flag at the end of `MoveGen` | +0.02% | [-0.21%, +0.36%] |
| fix, inline two-field test, no flag | **-1.44%** | [-1.79%, -1.05%] |
| flag at end vs flag after the bag count | -0.00% | [-0.18%, +0.17%] |

**Verdict.** On the current base the fix is indistinguishable from `main`, and the layout shift contributes nothing (the two flag placements are within 0.2% of each other either way). The cheaper test is visible in the instruction counts but below the noise floor of this workload now that #614's record loop is specialized per generation; the +0.3% on the older loop was a codegen artifact of that loop, not a property of this change, and I would not expect it on another compiler. The one effect that does register is the alternative not taken: deciding `bag > 0 && sort == equity` per subrack costs **1.4%**, which is why the decision is made once per generation and stored as a flag.


jvc56 (approved):

PR #678, opened 2026-09-08
Prune wordmap candidates with positional word-plus-floater masks

olaugh:
## What this does

WIT now records letters **outside each occurrence of the fixed board word**. Prefix/suffix unions preserve repeated letters elsewhere in a completion, while removing letters supplied only by the board. This tightens shadow pruning and wordmap subrack filtering.

WordPlusFloater (WPF) adds a positional constraint for anchors that play through multiple blocks: a base word, another fixed letter at a signed offset, and the final word length. Each cell contains the union of remaining letters across every matching dictionary completion. Move generation selects the longest indexed base and intersects masks for all fixed letters outside it, including letters belonging to another word in the lane. Normal word and placement checks still decide which moves are legal.

WPF covers **every dictionary base of lengths two through four**. CSW24 has 7,140 such bases. Dense rows use the WIT's value IDs directly; there are no frequency lists, cutoff options or row maps. The immutable table is owned by WIT and shared by workers.

The native C maker follows the WMP/RIT conversion path. One command, `convert kwg2wit CSW24`, builds ordinary and positional masks from the KWG and publishes a version-4 `.wit` file atomically. Both use the same terminal IDs and KWG fingerprint. No separate text lexicon, frequency list or Python runtime is needed. [Generation instructions](https://github.com/jvc56/MAGPIE/blob/perf-word-plus-floater/tools/WPF.md).

Version-3 WITs remain readable with ordinary filtering. Regenerate an existing WIT before training PGO to enable the full positional section. Alphabets beyond the positional encoding retain existing ordinary WIT behavior where supported. There is one file to generate and distribute.

## Performance

**Whole-stack comparison:** native stack #681 (#678 → #679 → #680), at `cf0fbb37` versus main `89f54e0c`, gives **+6.04% geometric throughput, +4.30% paired median** across 16 retained pairs. [#680 contains the complete measurement, variability and profile results](https://github.com/jvc56/MAGPIE/pull/680). This measures all three PRs together and does not isolate this PR’s gain.

Measured `3d111e13` against `main` at `89f54e0c`: Apple M4 Mac mini, CSW24, four workers, WMP/WIT and RIT mmap enabled, two plies, 0.5 seconds per simulated turn. Both sources received their own fresh production PGO profile from 16 static games with four configured training workers, using Apple Clang 17 and `make release`.

**16 fresh seed pairs**, eight in each run order, with all 32 timed runs retained. Games play the top static-equity move to hold the board trajectory fixed while measuring simulation iterations per second. Configuration and table loading are outside the timed region. Both binaries use the same temporary seedable benchmark harness; the submitted harness is unchanged.

| Configuration | Geometric mean gain | Nominal paired 95% CI | Paired median |
|---|---:|---:|---:|
| Production PGO, RIT on | **+2.60%** | **[+1.51%, +3.71%]** | **+2.12%** |

**15/16 pairs favor this change.** The interval is a Student-t interval over paired log throughput ratios. This measures residual WIT and full WPF together. It describes this host, workload and pair of trained binaries; it does not establish gains on other hardware or across repeated PGO training.

Candidate-first pairs average +2.97% and candidate-second pairs +2.24%; these are descriptive order summaries, with no adjustment or exclusions. Boundary snapshots show background activity; continuous clock and thermal telemetry was not collected. Balanced order does not eliminate transient shared-host noise.

The positional masks use **97.32 MiB of shared memory**, excluding allocator overhead. The complete combined CSW24 `.wit` is 121,972,344 bytes. The final format consolidation and generator checks preserve every numeric mask and the move-generation algorithm from the measured snapshot; the final packaged binary has not been separately compared against main.

Six macOS captures compare the frozen PGO binaries on three separate seeds, using a six-second window after one second of startup. Pooled exclusive nonwaiting **simulation-worker** shares were:

| Function/path | Main | This PR |
|---|---:|---:|
| `generate_moves` self, including inlined work | 37.22% | 36.88% |
| Named shadow functions | 26.04% | 25.94% |
| `wordmap_gen_forbidden_subracks` | 2.84% | 1.79% |
| `wmp_entry_write_blanks_to_buffer` | 5.12% | 5.37% |
| `record_wmp_plays_for_word` | 3.08% | 3.01% |

`wmp_get_word_entry` appears separately at 0.77% in the candidate; it is not separately resolved in main. The code paths affected are:

| Function/path | Effect |
|---|---|
| `wit_shadow_right_block_dead` | Residual WIT masks can stop shadow continuations earlier. |
| `wmp_move_gen_set_playthrough_bit_rack` | Intersects positional masks while preparing an anchor's fixed letters. |
| `wordmap_gen` / `wordmap_gen_forbidden_subracks` | Rejects anchors and subracks using the tighter letter set. |
| `wordmap_gen_record_subrack` → `wmp_move_gen_get_subrack_words` / `wmp_get_word_entry` | Downstream lookup and materialization that earlier filtering can avoid. |
| `wordmap_gen_check_playthrough_and_crosses` | Retains the normal placement checks for surviving words. |

Sample shares locate the affected path; they are not per-function speedups or avoided-call counts. PGO inlining folds work into callers such as `generate_moves`, and fixed windows can cover different game phases. Unchanged leave/exchange lookup accounts for 13.33%/13.62% of the same samples. These profiles do not separate WIT's gain from WPF's; the paired throughput comparison measures the net effect.

## Correctness

- Production-PGO, optimized and ASan/UBSan builds pass complete WIT-on/off move comparisons across the same **32 seeded CSW24 games / 723 positions**, including move identities, scores and equities before and after copy/undo.
- The native generator reproduces all **25,512,604 full-table masks** from the reference table exactly, including after the final generator fixes. C maker tests compare every cell of a small dictionary with a literal-position oracle, including repeated occurrences and board-edge offsets.
- Native generator, dense loader and synthetic positional-filter checks pass explicitly with `BOARD_DIM=21` under ASan/UBSan.
- Tests cover repeated letters, designated blanks, both signed offsets, all fixed letters in other blocks, longest-base selection in both scan orders and foreign cached rows. Combined-file tests cover v3 fallback, present zero masks versus absent sections, unknown flags, malformed trie structure, truncation, trailing bytes, cleanup and preservation of existing output on write failure.
- Formatting, circular dependencies and targeted cppcheck pass. Native clang-tidy findings were checked against main; see the PR checks for full Linux CI.

No persistent board or undo state is added. Benchmark score agreement is a workload check, separate from the complete-move correctness comparisons.

## What can build on this

The tighter mask is available before word lookup and recording. Deferring suitable eager subrack preparation until after that decision is a focused follow-up, subject to complete-move validation and fresh paired PGO measurement.


src/ent/word_info_table.c
```diff
@@ -0,0 +1,393 @@
...
+  for (size_t index = 0; index < count; index++) {
+    const uint32_t value = values[index];
+    const uint8_t bytes[4] = {(uint8_t)value, (uint8_t)(value >> 8),
+                              (uint8_t)(value >> 16), (uint8_t)(value >> 24)};
+    if (fwrite(bytes, sizeof(bytes), 1, stream) != 1) {
+      return false;
+    }
+  }
+  return true;
+#endif
```
jvc56: Let's try to move endian dependent code into compat so other code doesn't have to check for IS_LITTLE_ENDIAN.

src/ent/word_info_table.c
```diff
@@ -0,0 +1,393 @@
...
+  }
+  wit_clear(wit);
+  FILE *stream = stream_from_filename(filename, error_stack);
+  if (!error_stack_is_empty(error_stack)) {
+    return;
+  }
+  const char *message = "invalid or truncated word info table";
+  error_code_t status = ERROR_STATUS_RW_READ_ERROR;
+  if (fseek(stream, 0, SEEK_END) != 0) {
+    goto invalid;
```
jvc56: We should [avoid using goto statements](https://homepages.cwi.nl/~storm/teaching/reader/Dijkstra68.pdf) unless absolutely necessary or if it is a third party plugin like linenoise that we didn't write ourselves.

jvc56 (changes requested):

olaugh:
I’ll replace the `goto` cleanup paths with structured control flow throughout the WPF stack, preserving failure cleanup and loader validation. I’ll also open a separate PR against main to remove its existing C `goto` statements and add an explicit prohibition to AGENTS.md.

src/ent/word_info_table.c
olaugh: Fixed in 2a833d8be48e26a8e3d60a0c4ecd3cb9e2ac3dad. The loader now uses a parsing helper with early returns; word_info_table_load owns the stream and clears partial allocations on failure. The cleanup is propagated through the WPF stack. The WIT/conversion/WPF/positional-mask tests pass under ASan/UBSan. The separate main cleanup and AGENTS.md rule are in #683.

src/ent/word_info_table.c
olaugh: Fixed in 07a9c560. The shared fread_le_uint32s/fwrite_le_uint32s helpers now live in src/compat/endian_io.h and use the existing le32toh/htole32 conversions. word_info_table.c no longer checks IS_LITTLE_ENDIAN. The bulk little-endian path is preserved. Fixed wire-byte, empty-array, and short-read checks plus the WIT/WPF/conversion/positional-mask suites pass under ASan/UBSan. The change has been propagated through the native stack.

jvc56 (approved):

